PR #5, opened 2016-05-19
An example of connecting actor system's event bus to an SSE source

zunder:
No merging please. Just an example for now.

@aquamatthias the example is working now.

You can connect to the server through `curl localhost:9000/connect`. By accessing `localhost:9000/postmessage`, new messages appear in the terminal.


aquamatthias:
Thanks! Serves as example.

PR #27, opened 2016-06-10
Implemented scratch version of REST log setting

skindiakov:


aquamatthias:
Thanks - we will consider this for the next release.

PR #34, opened 2016-06-17
Validate incoming json with json schema.

aquamatthias:


api/src/main/resources/public/api/schema/v0/jobspec_v0.schema.json
```diff
@@ -0,0 +1,239 @@
...
+                    "type": "string",
+                    "description": "The operator for this constraint.",
+                    "enum": ["EQ", "LIKE", "UNLIKE"]
+                  },
+                  "value": {
+                    "type": "string",
+                    "description": "The value for this constraint."
+                  }
+                },
+                "required": ["attribute", "operator"]
```
orientzc: sorry.  i am looking for the ability to restrict a job to run a dedicated agent (like UNIQUE:hostname in marathon). and i have tried 
"placement": {
      "constraints": [
        {
          "attribute": "hostname",
          "operator": "unique"
        }
      ]
    },
or
"placement": {
      "constraints": [
        {
          "attribute": "hostname",
          "operator": "UNIQUE"
        }
      ]
    },
but it doesn't work.i think opreation doesn't support "unique" or "UINIQUE", it suppport "EQ", "LIKE", "UNLIKE", yes?
could you help me?

orientzc: sorry. i  am looking for the ability to restrict a job to run a dedicated agent (like UNIQUE:hostname in marathon). and i try
"placement": {
      "constraints": [
        {
          "attribute": "hostname",
          "operator": "unique"
        }
      ]
    },
or 
"placement": {
      "constraints": [
        {
          "attribute": "hostname",
          "operator": "UNIQUE"
        }
      ]
    },
but it doesn't work. what do i miss?
i think operation doesn't support "unique" or "UNIQUE", it support "EQ", "LIKE", "UNLIKE". yes? so what can i do? Can anybody help me?

meichstedt: @orientzc restricting ("pinning") a job to an agent is done by applying e.g. `hostname LIKE my-host` – more or less the same as in Marathon btw.

the `UNIQUE` operator doesn't apply in metronome because it can't schedule multiple instances at once – the `UNIQUE` semantic is _schedule n instances to n unique hosts_.

orientzc: @meichstedt thank you very very much. and i see with your reply. thank you. but i am sorry that  i am looking for he ability to use "unique" for the goal. the goal is that a new job will not launch if there is an existing run, and the new job will launch if there is not an existiong run or after the previous job finishes. and i try "concurrencyPolicy", but it is only support “ALLOW”. 
could you help me? thanks very very much.

orientzc: a new job will not launch if there is an existing run, and the new job will launch if there is not an existiong run or after the previous job finishes. it is to avoid to operate the table of the database at the same time.

orientzc: sorry.  i find it that "constraints",like this:
"placement": {
"constraints": [
{
"attribute": "hostname",
"operator": "EQ",
"value": "x.x.x.x"
}
]
},
but it not support public node, it only support private node. so what can i do? Can anybody help me?

aquamatthias: To run tasks on the public agents, a special mesos role `slave_public` is needed. metronome currently does not have the ability to specify accepted resource roles.

orientzc: @aquamatthias thank you.  and could you help me to understand the "concurrencyPolicy". it is only support “ALLOW”. but i am looking for the ability that  a new job will not launch if there is an existing run, and the new job will launch if there is not an existiong run or after the previous job finishes. it is to avoid to operate the table of the database at the same time. also, i cannot find this solution to solve it. could you help me? thanks very much.

PR #78, opened 2016-07-18
Fixes DCOS-8233 by validating the id in the json and the query param.

aquamatthias:


meichstedt:
I think the version in #77 is ok.

PR #88, opened 2016-08-01
Fixes DCOS-8909 by deleting the history entry for all job runs, if the job spec is deleted.

aquamatthias:
See https://mesosphere.atlassian.net/browse/DCOS-8909


unterstein:
Looks good, merged :-)

PR #91, opened 2016-08-10
version update to 1.9-snapshot

kensipe:
the version on master should be greater than the current release.  updating to 1.9


unterstein:
makes totally sense, merged (=

PR #92, opened 2016-08-16
Fixed ArtifactFormat

meichstedt:
* the artifact format now correctly parses `uri` as a required parameter, as stated in the schema

### DIFF
--- a/api/src/main/scala/dcos/metronome/api/v1/models/package.scala
+++ b/api/src/main/scala/dcos/metronome/api/v1/models/package.scala
@@ -35,7 +35,7 @@ package object models {
   )
 
   implicit lazy val ArtifactFormat: Format[Artifact] = (
-    (__ \ "url").format[String] ~
+    (__ \ "uri").format[String] ~
     (__ \ "extract").formatNullable[Boolean].withDefault(true) ~
     (__ \ "executable").formatNullable[Boolean].withDefault(false) ~
     (__ \ "cache").formatNullable[Boolean].withDefault(false)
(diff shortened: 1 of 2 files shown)


kensipe:
@meichstedt there is a test failure.  I double checked... master is good.   This change breaks: dcos.metronome.jobrun.impl.JobRunExecutorActorTest

kensipe:
@meichstedt this is a simple change with a test.  It LGTM as long as the other tests are also working.

api/src/main/scala/dcos/metronome/api/v1/models/package.scala
```diff
@@ -35,7 +35,7 @@ package object models {
   )
 
   implicit lazy val ArtifactFormat: Format[Artifact] = (
-    (__ \ "url").format[String] ~
+    (__ \ "uri").format[String] ~
```
gkleiman: The design doc specifies `url`, and according to https://danielmiessler.com/study/url-uri/ (I admit it might not be the best source, but I'm lazy today :P) it is a URL, which in my opinion is a more user-friendly/known name.

But I'd be ok with calling this `uri` to make it more consistent with Marathon's API...

kensipe:
@meichstedt I double checked and the testing failure was only on my machine.  This looks great!  I'll merge.

api/src/main/scala/dcos/metronome/api/v1/models/package.scala
meichstedt: Yet the [schema](https://github.com/dcos/metronome/blob/master/api/src/main/resources/public/api/v1/schema/jobspec.schema.json#L48) specifies the property to be `uri`, and imo the released schema is the source of truth in this case. I'd be totally ok with renaming it to `url` to be more specific about the intent, but this would be a breaking change.

PR #102, opened 2016-11-07
API Examples update: test.com -> example.com

lloesche:
The domain test.com is actually owned by a company. In accordance with https://tools.ietf.org/html/rfc2606#section-3 we should be using example.com instead as it's reserved for this purpose.


aquamatthias:
Thanks!

PR #104, opened 2016-11-09
si tests

kensipe:


tests/system/metronome.py
```diff
@@ -0,0 +1,550 @@
...
+        :type resource_id: str
+        :param resource_json: the json payload
+        :type resource_json: {}
+        :param force: whether to override running deployments
+        :type force: bool
+        :returns: the response from Metronome
+        :rtype: requests.Response
+        """
+
+        path_template = 'v1/{}/{{}}'.format(resource_type)
```
colin-msphere: Two `{}` instances with only one item in `format`.

tests/system/metronome.py
```diff
@@ -0,0 +1,550 @@
...
+
+    def update_schedule(self, job_id, schedule_id, schedule_resource):
+        """Gets the schedules for a given job
+
+        :param job_id: the ID of the job to remove
+        :type job_id: str
+        :rtype: json of schedules
+        """
+        job_id = util.normalize_marathon_id_path(job_id)
+        if hasattr(schedule_resource, 'read'):
```
colin-msphere: I have seen this block enough that it should be moved to a method.

PR #119, opened 2017-01-19
Fix typo in schedulespec schema

mindscratch:
Instead of `FORBID` it had `FORBIF`.


aquamatthias:
@mindscratch Thanks! Please see #121 - this will remove that part of the sentence.

PR #105, opened 2016-11-15
Fixes #100 | Added ALLOW and FORBID to proto

janisz:

PR #121, opened 2017-01-26
Remove FORBID and REPLACE from the code as well as documentation (Fixes #100)

### LINKED ISSUE #100: Unable to create a schedule with concurrencyPolicy = FORBID on DC/OS 1.8.4
I'm unable to create a schedule which doesn't start a new job before the old one has finished.

I'm using the REST API of DC/OS 1.8.4 as follows:
`POST /service/metronome/v1/jobs/<my-job>/schedules`

```
{
    "id": "default",
    "enabled": true,
    "cron": "*/5 * * * *",
    "concurrencyPolicy": "FORBID"
}
```

But I'm getting the following error:
`{"message":"No enum constant dcos.metronome.Protos.JobSpec.ScheduleSpec.ConcurrencyPolicy.FORBID"}`

What am I missing?

aquamatthias:
The default value is set to ALLOW.
The schema will make sure, only ALLOW is allowed.

### DIFF
--- a/api/src/main/resources/public/api/v1/schema/schedulespec.schema.json
+++ b/api/src/main/resources/public/api/v1/schema/schedulespec.schema.json
@@ -30,7 +30,7 @@
     "concurrencyPolicy": {
       "type": "string",
       "enum": ["ALLOW"],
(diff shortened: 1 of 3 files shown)


aquamatthias:
Close in favor of #105

meichstedt:
Thanks!

PR #133, opened 2017-07-14
Upgrade cron-utils to 4.1.1

dougkoch:
There is a bug in cron-utils 4.1.0 which sometimes causes cron jobs that are set to run on a specific day of the week to have the wrong execution time if the intended execution day happens to fall on either the first or last day of the month. I have included a test which fails with cron-utils 4.1.0, but passes with 4.1.1.


jdef:
ok to test

kensipe:
there are lots of issues with cronutils!

kensipe:
I've added the tests you provided as we upgraded to cron-utils 6.0.3.   Thanks!!!
https://github.com/dcos/metronome/pull/137/files#diff-162c67d2a81738c20905de88aeada28bR142

PR #135, opened 2017-09-27
Adding a job example with placement

kensipe:
thanks to community member Aleksandar Simic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,41 @@ use. Note that you still need to run ZooKeeper for storing state. The following
 command launches Metronome on Mesos in *local mode*. 
 
     ./bin/metronome -Dmetronome.mesos.master.url=local
-
+    
+    
+ ## Example Job with Placement Constraint
+ 
+ ```
+ {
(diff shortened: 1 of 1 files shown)


dotemacs:
GitHub swallowed the *host-name* as it was in angle brackets...

README.md
```diff
@@ -45,13 +45,46 @@ To build Metronome from source, check out this repo and use sbt to build a unive
...
+         {
+             "id": "sample-schedule",
+             "enabled": true,
+             "cron": "0 0 * * *",
+             "concurrencyPolicy": "ALLOW"
+         }
+     ]
+}
+ ```
+ This job will sleep every day at midnight and will land on the host defined by `<host-name>` which could be the hostname or IP of a node in the cluster.  If you don't care where it lands in the cluster remove the `placement
```
dotemacs: Sorry to be a pain, but you've missed out the closing backtick for `placement`.

kensipe: turns out all the things matter.. thx!!

PR #140, opened 2017-10-23
Define Metronome build pipeline in Jenkinsfile.

jeschkies:


alenkacz:
in general we are moving everything from travis to jenkins?

jeschkies:
> in general we are moving everything from travis to jenkins?

Seems like it.

PR #147, opened 2017-11-28
Cron Min Reschedule is 60s

kensipe:
adding 60 sec min time for schedules

https://jira.mesosphere.com/browse/MARATHON_EE-1717
JIRA: MARATHON_EE-1717


kensipe:
@alenkacz I'm not clear on your comment.   I believe it says 60 secs in all places.   perhaps I made a mistake and it was edited away.  sorry about that if that was the case.

PR #150, opened 2017-12-01
Adding version info to metronome

kensipe:


kensipe:
@timcharper I still can't add you as a reviewer but it would be awesome if you have time.

api/src/main/scala/dcos/metronome/api/v1/controllers/ApplicationController.scala
```diff
@@ -19,6 +21,18 @@ class ApplicationController(metrics: Metrics) extends RestController {
 
   def ping = Action { Ok("pong") }
 
+  def info = Action {
+
+    val infoJson = Json.toJson(
```
alenkacz: maybe we should introduce a case class for this and serialize it using some marshaller ... ?

jobs/src/main/scala/dcos/metronome/MetronomeBuildInfo.scala
```diff
@@ -0,0 +1,67 @@
...
+    }
+    version
+  }
+
+  /**
+    * sbt-native-package provides all of the files as individual JARs. By default, `getResourceAsStream` returns the
+    * first matching file for the first JAR in the class path. Instead, we need to enumerate through all of the
+    * manifests, and find the one that applies to the Metronome application jar.
+    */
+  lazy val ManifestPath: List[java.net.URL] =
```
alenkacz: this should start with lowercase probably...

api/src/main/scala/dcos/metronome/api/v1/controllers/ApplicationController.scala
kensipe: I don't see the value... less code is better code.

kensipe:
the raml was a big miss.. thx

api/src/main/scala/dcos/metronome/api/v1/controllers/ApplicationController.scala
```diff
@@ -19,6 +21,18 @@ class ApplicationController(metrics: Metrics) extends RestController {
...
+      Map(
+        "version" -> MetronomeBuildInfo.version,
+        "lib-version" -> MetronomeBuildInfo.marathonVersion
```
alenkacz: do we wanna put dash in the property name? I think so far we were doing camel casing... ?

btw. this would not happen if we had this encoded as case class as we do for all other models... :-)

kensipe: you're right!

jobs/src/main/scala/dcos/metronome/MetronomeBuildInfo.scala
```diff
@@ -0,0 +1,70 @@
...
+import scala.util.control.NonFatal
+import scala.util.{ Failure, Success, Try }
+
+/**
+  * Provides build information details regarding Metronome and Marathon at runtime.
+  */
+case object MetronomeBuildInfo {
+  private val metronomeJar = "\\bmesosphere\\.metronome\\.metronome-[0-9.]+".r
+
+  private lazy val devBuildVersion = {
```
timcharper: heh this is such a hack :)

timcharper: Probably out of the scope for this change, and I know how much you love messing with sbt :) but it would be really great if sbt would write out the manifest file to a resource path on build, so that we could always just read the manifest.

api/src/main/scala/dcos/metronome/api/v1/controllers/ApplicationController.scala
```diff
@@ -19,6 +21,10 @@ class ApplicationController(metrics: Metrics) extends RestController {
 
   def ping = Action { Ok("pong") }
 
+  def info = Action {
+    Ok(MetronomeInfoWrites.writes(MetronomeInfoBuilder.metronomeInfo))
```
meichstedt: The writes is implicit, you shouldn't need to explicitly use it here.

(In case you did that to make it obvious where the conversion comes from, that's fine with me.)

jobs/src/main/scala/dcos/metronome/MetronomeBuildInfo.scala
meichstedt: +1

api/src/main/scala/dcos/metronome/api/v1/controllers/ApplicationController.scala
alenkacz: yeah we talked with Ken about this use of implicits and we agreed that we prefer doing these explicitly...

meichstedt: That's what I imagined :)

jobs/src/main/scala/dcos/metronome/MetronomeBuildInfo.scala
kensipe: yeah I thought about that.  I'll add a jira to do it.    It isn't as bad of a hack as Marathon IMO.   The default there is no good IMO.  I would rather have the manifest file as well.

kensipe: https://jira.mesosphere.com/browse/DCOS_OSS-1939

PR #151, opened 2017-12-07
Make it possible to enable framework authentication in metronome

alenkacz:
Summary: We updated marathon inside metronome from 1.2.0-RC to 1.3.13. 1.3.0 version of marathon introduced new mesos authentication flag (see https://github.com/mesosphere/marathon/releases/tag/v1.3.0). Without that flag, authentication is not enabled even if principal is provided.
Currently, dcos integration tests enable authentication through env. variables like this https://github.com/mesosphere/dcos-enterprise/blob/master/gen_extra/dcos-config.yaml#L845
I have added new flag to metronome that enables authentication explicitly.
After this patch is merged, the next step is to patch dcos-config.yaml to pass this env. variable that will enable the authentication again.

For more information see this pull request https://github.com/mesosphere/dcos-enterprise/pull/1813

JIRA: MARATHON_EE-1726

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -386,7 +386,7 @@ metronome {
 
   mesos {
     #  (Optional Default: false) Enable framework authentication.
-    authentication = ${?METRONOME_MESOS_AUTHENTICATION}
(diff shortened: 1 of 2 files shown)


kensipe:
nice work!

PR #153, opened 2017-12-11
Play version should be the same for the plugin

alenkacz:
I think this two should be in sync. And we updated play in one of the previous commits so we should update this too...


alenkacz:
I did not find a way to share that version between those two files :( Anyone have an idea? @timcharper 

This is where I ended up... https://stackoverflow.com/questions/23944108/how-to-share-version-values-between-project-plugins-sbt-and-project-build-scala

PR #156, opened 2017-12-12
Add taskKillGracePeriod

kensipe:
Added ability to send taskKillGracePeriodSeconds to marathon libs from metronome.
Changed the name to include suffix `Seconds` which is consistent with parts of marathon:)

https://jira.mesosphere.com/browse/METRONOME-151

Jira:  METRONOME-151

### DIFF
--- a/jobs/src/main/protobuf/metronome.proto
+++ b/jobs/src/main/protobuf/metronome.proto
@@ -173,6 +173,14 @@ message JobSpec {
       optional int64 activeDeadline = 2;
     }
     optional RestartSpec restart = 13;
+
+    // mesos kill policy https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L625
+    // for some reason referred to as taskKillGracePeriod in marathon.
+    // https://github.com/mesosphere/marathon/blob/releases/1.3/src/main/scala/mesosphere/marathon/state/AppDefinition.scala#L79
+    // AND taskKillGracePeriodSeconds
+    // https://github.com/mesosphere/marathon/blob/releases/1.3/docs/docs/rest-api/public/api/v2/schema/AppDefinition.json
+    // int64 in mesos and int32 in marathon... going with int64
+    optional int64 taskKillGracePeriodSeconds = 14;
   }
   optional RunSpec run = 6;
 }
--- a/jobs/src/main/protobuf/metronome.proto
+++ b/jobs/src/main/protobuf/metronome.proto
@@ -180,7 +180,7 @@ message JobSpec {
     // AND taskKillGracePeriodSeconds
     // https://github.com/mesosphere/marathon/blob/releases/1.3/docs/docs/rest-api/public/api/v2/schema/AppDefinition.json
     // int64 in mesos and int32 in marathon... going with int64
-    optional int64 taskKillGracePeriodSeconds = 14;
+    optional int64 TaskKillGracePeriodSeconds = 14;
   }
   optional RunSpec run = 6;
 }
--- a/jobs/src/main/protobuf/metronome.proto
+++ b/jobs/src/main/protobuf/metronome.proto
@@ -180,7 +180,7 @@ message JobSpec {
     // AND taskKillGracePeriodSeconds
     // https://github.com/mesosphere/marathon/blob/releases/1.3/docs/docs/rest-api/public/api/v2/schema/AppDefinition.json
     // int64 in mesos and int32 in marathon... going with int64
-    optional int64 TaskKillGracePeriodSeconds = 14;
+    optional int64 task_kill_grace_period_seconds = 14;
   }
   optional RunSpec run = 6;
 }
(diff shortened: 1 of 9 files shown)


kensipe:
Looking into what it takes... manually tested.. needs unit tests

jobs/src/main/protobuf/metronome.proto
```diff
@@ -173,6 +173,14 @@ message JobSpec {
...
     }
     optional RestartSpec restart = 13;
+
+    // mesos kill policy https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L625
+    // for some reason referred to as taskKillGracePeriod in marathon.
+    // https://github.com/mesosphere/marathon/blob/releases/1.3/src/main/scala/mesosphere/marathon/state/AppDefinition.scala#L79
+    // AND taskKillGracePeriodSeconds
+    // https://github.com/mesosphere/marathon/blob/releases/1.3/docs/docs/rest-api/public/api/v2/schema/AppDefinition.json
+    // int64 in mesos and int32 in marathon... going with int64
+    optional int64 taskKillGracePeriodSeconds = 14;
```
meichstedt: I'd stick with int32 – Mesos' `DurationInfo` specifies the amount in nanoseconds, that's why it's an int64.

jobs/src/main/scala/dcos/metronome/repository/impl/kv/marshaller/JobSpecMarshaller.scala
```diff
@@ -136,6 +136,7 @@ object RunSpecConversions {
       runSpec.args.foreach { args => builder.addAllArguments(args.asJava) }
       runSpec.user.foreach(builder.setUser)
       runSpec.docker.foreach { docker => builder.setDocker(docker.toProto) }
+      runSpec.taskKillGracePeriodSeconds.foreach { killGracePeriod => builder.setTaskKillGracePeriodSeconds(killGracePeriod.toMillis) }
```
meichstedt: Wait a _second_ – shouldn't this be **`toSeconds`**? 🙄 

Also, you're missing the conversion in `implicit class ProtoToRunSpec` on line 145ff.

(In the current impl the gracePeriod would be received via the API, stored in zk and passed over to Marathon, but not retrieved from zk after a failover).

jobs/src/main/protobuf/metronome.proto
alenkacz: also I think that the code style for proto files require to use _ not camelCase. We should stick with that...

kensipe: all of our protobuf defs are camel... however I should have started with an upper case.

jobs/src/main/protobuf/metronome.proto
```diff
@@ -173,6 +173,14 @@ message JobSpec {
...
     }
     optional RestartSpec restart = 13;
+
+    // mesos kill policy https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L625
+    // for some reason referred to as taskKillGracePeriod in marathon.
+    // https://github.com/mesosphere/marathon/blob/releases/1.3/src/main/scala/mesosphere/marathon/state/AppDefinition.scala#L79
+    // AND taskKillGracePeriodSeconds
+    // https://github.com/mesosphere/marathon/blob/releases/1.3/docs/docs/rest-api/public/api/v2/schema/AppDefinition.json
+    // int64 in mesos and int32 in marathon... going with int64
+    optional int64 TaskKillGracePeriodSeconds = 14;
```
alenkacz: oh well, now it is a weird mix, there is max_launch_delay but also activeDeadline ... I would personally stick with the official guidelines which is task_kill_grace...

kensipe: aw.. you're right

alenkacz:
I think we should test:
- it is marshalled to proto (that one I already commited)
- it is properly parsed in controller
- it is passed to marathon (this one I am working on)

I think we don't need integration/SI test ... we can verify it is passed to marathon and we can expect that this is already tested in integration test in Marathon ... ? :)

I have a "JobRunExecutorActor" test already halfway ready - I will commit that one tomorrow.

Otherwise nice work @kensipe !

alenkacz:
@kensipe I have submitted the tests so feel free to review if it makes sense to you like this

jobs/src/main/protobuf/metronome.proto
```diff
@@ -173,6 +173,14 @@ message JobSpec {
...
     }
     optional RestartSpec restart = 13;
+
+    // mesos kill policy https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L625
+    // for some reason referred to as taskKillGracePeriod in marathon.
+    // https://github.com/mesosphere/marathon/blob/releases/1.3/src/main/scala/mesosphere/marathon/state/AppDefinition.scala#L79
+    // AND taskKillGracePeriodSeconds
+    // https://github.com/mesosphere/marathon/blob/releases/1.3/docs/docs/rest-api/public/api/v2/schema/AppDefinition.json
+    // int64 in mesos and int32 in marathon... going with int64
+    optional int64 task_kill_grace_period_seconds = 14;
```
alenkacz: Matthias proposed to stick with int32 here. Not sure whether his comment was just lost or whether it was a decision to keep int64 so just pointing it out...

kensipe: int32 adds unnecessary complicity... we had a slack conversation which resulted in being ok to use int64

PR #159, opened 2017-12-15
HTTP 404 error with a proper message

alenkacz:
Summary: When no route handler is found for a given route, we currently return HTTP 404 with no message. This PR adds a message.

This is a recommended way to do so, see GlobalSettings.onHandlerNotFound here https://www.playframework.com/documentation/2.6.x/GlobalSettings

JIRA: METRONOME-91

### DIFF
--- a/api/src/main/scala/dcos/metronome/api/ErrorHandler.scala
+++ b/api/src/main/scala/dcos/metronome/api/ErrorHandler.scala
@@ -37,5 +37,5 @@ class ErrorHandler extends HttpErrorHandler {
 }
 
 object ErrorHandler {
-  val noRouteHandlerMessage = "Unsupported route, no handler found."
+  val noRouteHandlerMessage = "Unknown route."
 }
--- a/api/src/main/scala/dcos/metronome/api/ErrorHandler.scala
+++ b/api/src/main/scala/dcos/metronome/api/ErrorHandler.scala
@@ -15,7 +15,8 @@ class ErrorHandler extends HttpErrorHandler {
 
   override def onClientError(request: RequestHeader, statusCode: Int, message: String): Future[Result] = {
     log.debug(s"Client Error on path ${request.path}. Message: $message Status: $statusCode")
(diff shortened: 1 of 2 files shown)


api/src/main/scala/dcos/metronome/api/ErrorHandler.scala
```diff
@@ -34,3 +35,7 @@ class ErrorHandler extends HttpErrorHandler {
     */
   private def escape(msg: String): String = HtmlFormat.escape(msg).body
 }
+
+object ErrorHandler {
+  val noRouteHandlerMessage = "Unsupported route, no handler found."
```
kensipe: lets change the message here...   perhaps "Unknown route." is sufficient

reasoning:
1. it seems direct and to the point and is what a 404 is.   
2. I don't like exposing internal details to an http client...  "no handler found" seems like an internal concern.

kensipe: otherwise looks good

meichstedt: "Unknown route." sounds sufficient to me as well

api/src/test/scala/dcos/metronome/api/v1/controllers/JobSpecControllerTest.scala
```diff
@@ -1,10 +1,10 @@
 package dcos.metronome.api.v1.controllers
 
 import dcos.metronome.api.v1.models._
-import dcos.metronome.api.{ MockApiComponents, OneAppPerTestWithComponents, TestAuthFixture, UnknownJob }
+import dcos.metronome.api._
```
meichstedt: Wildcard import? 😱 
Well in test code it might be OK – IIRC our conventions are: No?

alenkacz: :) fixed

alenkacz:
@kensipe message fixed

PR #163, opened 2017-12-21
Do not compile protoc on every build

alenkacz:
Summary:
I don't think we need to compile protobuf from sources on every jenkins run. Google publishes already precompiled binaries for all platforms.

This PR alters the install script in a way that it takes advantage of these precompiled binaries.

JIRA issues: METRONOME-188

### DIFF
--- a/bin/install-protobuf.sh
+++ b/bin/install-protobuf.sh
@@ -10,9 +10,8 @@ fi
 if [ ! -d "$HOME/protoc" ]; then
   curl -L -o protoc-3.3.0.zip $download_url
   unzip -d protoc-3.3.0 -a protoc-3.3.0.zip
-  mv -v ./protoc-3.3.0 ~/protoc
+  mv -v ./protoc-3.3.0 ~/protobuf
   rm protoc-3.3.0.zip
-  export PATH=~/protoc/bin:$PATH
 else
   echo "Using already installed protoc."
 fi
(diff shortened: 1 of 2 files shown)


bin/install-protobuf.sh
```diff
@@ -1,10 +1,17 @@
...
+unamestr=`uname`
+echo $unamestr
+if [ "$unamestr" == 'Darwin' ]; then
+   download_url='https://github.com/google/protobuf/releases/download/v3.3.0/protoc-3.3.0-osx-x86_64.zip'
+fi
+if [ ! -d "$HOME/protobuf" ]; then
+  curl -L -o protoc-3.3.0.zip $download_url
+  unzip -d protoc-3.3.0 -a protoc-3.3.0.zip
+  mv -v ./protoc-3.3.0 ~/protobuf
+  rm protoc-3.3.0.zip
```
kensipe: 1 more change would make this cherry....    would it make sense to pull the version out as a variable so that it could be easily changed once.     var="3.3.0"

timcharper: Even better... make it an argument, and default it to 3.3.0?

Extra mile, put the version in a property file, source it in this script, and then update sbt to read the same property file?

alenkacz: good point, I made it parameter with default for now...

PR #164, opened 2017-12-27
Implement start deadline

alenkacz:
Summary:
This implements startingDeadline - a property that was part of API but never did anything. Now it should not launch a job that was not able to launch until this timeout is reached.

JIRA issues: METRONOME-191

### DIFF
--- a/jobs/src/main/scala/dcos/metronome/jobrun/JobRunModule.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobrun/JobRunModule.scala
@@ -28,10 +28,10 @@ class JobRunModule(
 
   import com.softwaremill.macwire._
 
-  private[this] def executorFactory(jobRun: JobRun, startingDeadline: Option[Duration], promise: Promise[JobResult]): Props = {
+  private[this] def executorFactory(jobRun: JobRun, promise: Promise[JobResult]): Props = {
     val persistenceActorFactory = (id: JobRunId, context: ActorContext) =>
       context.actorOf(JobRunPersistenceActor.props(id, jobRunRepository, behavior))
-    JobRunExecutorActor.props(jobRun, startingDeadline, promise, persistenceActorFactory,
+    JobRunExecutorActor.props(jobRun, promise, persistenceActorFactory,
       launchQueue, taskTracker, driverHolder, clock, behavior)
   }
 
--- a/jobs/src/main/scala/dcos/metronome/jobrun/JobRunModule.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobrun/JobRunModule.scala
@@ -32,7 +32,7 @@ class JobRunModule(
     val persistenceActorFactory = (id: JobRunId, context: ActorContext) =>
       context.actorOf(JobRunPersistenceActor.props(id, jobRunRepository, behavior))
     JobRunExecutorActor.props(jobRun, promise, persistenceActorFactory,
-      launchQueue, taskTracker, driverHolder, clock, behavior)
+      launchQueue, taskTracker, driverHolder, clock, behavior)(actorSystem.scheduler)
   }
 
   val jobRunServiceActor = leadershipModule.startWhenLeader(
--- a/jobs/src/main/scala/dcos/metronome/jobrun/JobRunModule.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobrun/JobRunModule.scala
@@ -1,10 +1,10 @@
 package dcos.metronome
 package jobrun
 
-import akka.actor.{ ActorContext, ActorSystem, Props }
+import akka.actor.{ActorContext, ActorSystem, Props}
 import dcos.metronome.behavior.Behavior
-import dcos.metronome.jobrun.impl.{ JobRunExecutorActor, JobRunPersistenceActor, JobRunServiceActor, JobRunServiceDelegate }
-import dcos.metronome.model.{ JobResult, JobRun, JobRunId }
+import dcos.metronome.jobrun.impl.{JobRunExecutorActor, JobRunPersistenceActor, JobRunServiceActor, JobRunServiceDelegate}
+import dcos.metronome.model.{JobResult, JobRun, JobRunId}
 import dcos.metronome.repository.Repository
 import dcos.metronome.utils.time.Clock
 import mesosphere.marathon.MarathonSchedulerDriverHolder
@@ -13,6 +13,7 @@ import mesosphere.marathon.core.task.tracker.TaskTracker
 import mesosphere.marathon.core.leadership.LeadershipModule
 
 import scala.concurrent.Promise
+import scala.concurrent.duration.Duration
 
 class JobRunModule(
   config:           JobRunConfig,
@@ -27,10 +28,10 @@ class JobRunModule(
 
   import com.softwaremill.macwire._
 
-  private[this] def executorFactory(jobRun: JobRun, promise: Promise[JobResult]): Props = {
+  private[this] def executorFactory(jobRun: JobRun, startingDeadline: Option[Duration], promise: Promise[JobResult]): Props = {
     val persistenceActorFactory = (id: JobRunId, context: ActorContext) =>
       context.actorOf(JobRunPersistenceActor.props(id, jobRunRepository, behavior))
-    JobRunExecutorActor.props(jobRun, promise, persistenceActorFactory,
+    JobRunExecutorActor.props(jobRun, startingDeadline, promise, persistenceActorFactory,
       launchQueue, taskTracker, driverHolder, clock, behavior)
   }
 
--- a/jobs/src/main/scala/dcos/metronome/jobrun/JobRunModule.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobrun/JobRunModule.scala
@@ -1,10 +1,10 @@
 package dcos.metronome
 package jobrun
 
-import akka.actor.{ActorContext, ActorSystem, Props}
+import akka.actor.{ ActorContext, ActorSystem, Props }
 import dcos.metronome.behavior.Behavior
-import dcos.metronome.jobrun.impl.{JobRunExecutorActor, JobRunPersistenceActor, JobRunServiceActor, JobRunServiceDelegate}
-import dcos.metronome.model.{JobResult, JobRun, JobRunId}
+import dcos.metronome.jobrun.impl.{ JobRunExecutorActor, JobRunPersistenceActor, JobRunServiceActor, JobRunServiceDelegate }
+import dcos.metronome.model.{ JobResult, JobRun, JobRunId }
 import dcos.metronome.repository.Repository
 import dcos.metronome.utils.time.Clock
 import mesosphere.marathon.MarathonSchedulerDriverHolder
--- a/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
@@ -1,12 +1,12 @@
 package dcos.metronome
 package jobrun.impl
 
-import akka.actor.{Actor, ActorContext, ActorLogging, ActorRef, Props, Stash}
-import dcos.metronome.{JobRunFailed, UnexpectedTaskState}
-import dcos.metronome.behavior.{ActorBehavior, Behavior}
+import akka.actor.{ Actor, ActorContext, ActorLogging, ActorRef, Cancellable, Props, Stash }
+import dcos.metronome.{ JobRunFailed, UnexpectedTaskState }
+import dcos.metronome.behavior.{ ActorBehavior, Behavior }
 import dcos.metronome.eventbus.TaskStateChangedEvent
 import dcos.metronome.jobrun.StartedJobRun
-import dcos.metronome.model.{JobResult, JobRun, JobRunId, JobRunStatus, JobRunTask, RestartPolicy}
+import dcos.metronome.model._
 import dcos.metronome.scheduler.TaskState
 import dcos.metronome.utils.time.Clock
 import mesosphere.marathon.MarathonSchedulerDriverHolder
@@ -38,6 +38,9 @@ class JobRunExecutorActor(
   import JobRunExecutorActor._
   import JobRunPersistenceActor._
   import TaskStates._
+  import context.dispatcher
+
+  private[impl] var startingDeadlineTimer: Option[Cancellable] = None
 
   lazy val persistenceActor = persistenceActorRefFactory(run.id, context)
   var jobRun: JobRun = run
@@ -46,7 +49,7 @@ class JobRunExecutorActor(
 
   override def preStart(): Unit = {
     scheduleStartingDeadlineTimeout()
-    
+
     jobRun.status match {
       case JobRunStatus.Initial => becomeCreating()
 
@@ -95,7 +98,7 @@ class JobRunExecutorActor(
       val now = clock.now()
       val from = run.createdAt.plus(deadline.toMillis)
       val timeout = Seconds.secondsBetween(now, from).getSeconds.seconds
-      context.system.scheduler.scheduleOnce(timeout, self, StartTimeout)
+      startingDeadlineTimer = Some(context.system.scheduler.scheduleOnce(timeout, self, StartTimeout))
     }
   }
 
@@ -106,12 +109,18 @@ class JobRunExecutorActor(
   }
 
   def becomeActive(update: TaskStateChangedEvent): Unit = {
+    cancelStartingDeadline()
     jobRun = jobRun.copy(status = JobRunStatus.Active, tasks = updatedTasks(update))
     persistenceActor ! Update(_ => jobRun)
     context.parent ! JobRunUpdate(StartedJobRun(jobRun, promise.future))
 
     log.debug("become active")
     context.become(active)
+  }
+
+  def cancelStartingDeadline(): Unit = {
+    startingDeadlineTimer.foreach { c => if (!c.isCancelled) c.cancel() }
+    startingDeadlineTimer = None
   }
 
   def updatedTasks(update: TaskStateChangedEvent): Map[Task.Id, JobRunTask] = {
--- a/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
@@ -26,7 +26,6 @@ import scala.concurrent.duration.Duration
   */
 class JobRunExecutorActor(
     run:                        JobRun,
-    startingDeadline:           Option[Duration],
     promise:                    Promise[JobResult],
     persistenceActorRefFactory: (JobRunId, ActorContext) => ActorRef,
     launchQueue:                LaunchQueue,
@@ -94,10 +93,11 @@ class JobRunExecutorActor(
   def scheduleTimeout(): Unit = {
     import scala.concurrent.duration._
 
-    startingDeadline.foreach { deadline =>
+    jobRun.startingDeadline.foreach { deadline =>
       val now = clock.now()
       val from = run.createdAt.plus(deadline.toMillis)
       val timeout = Seconds.secondsBetween(now, from).getSeconds.seconds
+      log.info(s"Timeout scheduled for ${jobRun.id} at $timeout")
       startingDeadlineTimer = Some(context.system.scheduler.scheduleOnce(timeout, self, StartTimeout))
     }
   }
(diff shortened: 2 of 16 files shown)


jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -41,6 +44,8 @@ class JobRunExecutorActor(
   val runSpecId = jobRun.id.toPathId
 
   override def preStart(): Unit = {
+    scheduleStartingDeadlineTimeout()
```
kensipe: personal preference... `scheduleTimeout(startingDeadline)`   I don't know why:) .  it is totally fine the way it is.

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -82,6 +87,17 @@ class JobRunExecutorActor(
...
   }
 
+  def scheduleStartingDeadlineTimeout(): Unit = {
+    import scala.concurrent.duration._
+
+    startingDeadline.foreach { deadline =>
+      val now = clock.now()
+      val from = run.createdAt.plus(deadline.toMillis)
+      val timeout = Seconds.secondsBetween(now, from).getSeconds.seconds
+      context.system.scheduler.scheduleOnce(timeout, self, StartTimeout)
```
kensipe: This has to be cancellable and needs to be cancelled on `start` :)
example:  https://github.com/dcos/metronome/blob/master/jobs/src/main/scala/dcos/metronome/jobspec/impl/JobSpecSchedulerActor.scala#L27
and
https://github.com/dcos/metronome/blob/master/jobs/src/main/scala/dcos/metronome/jobspec/impl/JobSpecSchedulerActor.scala#L73

kensipe: 1. I'm not sure `createdAt` is Some at this point...   is this not None for the first launch at preStart?
2. if createdAt is None, then use now for calculation
3. Love the use of createdAt... however the timeout could be negative if the time has past.   We should check for that and abort if true

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -207,14 +223,20 @@ class JobRunExecutorActor(
     case _ => stash()
   }
 
+  def receiveStartTimeout: Receive = {
+    case StartTimeout =>
+      log.info(s"Start timed out for JobRun ${jobRun.id} with deadline ${startingDeadline.get}")
```
kensipe: I would rather not deduce the the log... prefer "startingDeadline timeout expired..."  or something similar... "Start timed out..." isn't clear to me.   Along with jobRun.id and startingDeadline, it would be great to have createdAt.  Perhaps s"startingDeadline duration of ${startingDeadline.get} expired for JobRun ${jobRun.id} created at ${jobRun.createdAt}"

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -224,7 +246,7 @@ class JobRunExecutorActor(
   }
 
   def starting: Receive = around {
-    receiveKill orElse {
+    receiveKill orElse receiveStartTimeout orElse {
       case ForwardStatusUpdate(update) if isActive(update.taskState) =>
```
kensipe: seems like a good place to cancel the start timeout... or perhaps `active`

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunServiceActor.scala
```diff
@@ -47,7 +48,7 @@ class JobRunServiceActor(
     case KillJobRun(id)                => killJobRun(id)
 
     // trigger messages
-    case TriggerJobRun(spec)           => triggerJobRun(spec)
+    case TriggerJobRun(spec, startingDeadline)           => triggerJobRun(spec, startingDeadline)
 
     // executor messages
     case JobRunUpdate(started)         => updateJobRun(started)
```
kensipe: it would be great to re-align these `=> methods`

jobs/src/main/scala/dcos/metronome/jobrun/JobRunModule.scala
```diff
@@ -1,9 +1,9 @@
 package dcos.metronome.jobrun
 
-import akka.actor.{ ActorContext, ActorSystem, Props }
+import akka.actor.{ActorContext, ActorSystem, Props}
```
kensipe: apparently our formatter wants spaces here ... running compile reformats a bunch of code.

kensipe:
and the code doesn't compile 😢

kensipe:
[error] /Users/kensipe/projects/metronome/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala:97: Cannot find an implicit ExecutionContext. You might pass
[error] an (implicit ec: ExecutionContext) parameter to your method
[error] or import scala.concurrent.ExecutionContext.Implicits.global.
[error]       context.system.scheduler.scheduleOnce(timeout, self, StartTimeout)
[error]                                            ^
[error] /Users/kensipe/projects/metronome/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunServiceActor.scala:149: type mismatch;
[error]  found   : (dcos.metronome.model.JobRun, Option[scala.concurrent.duration.Duration]) => dcos.metronome.jobrun.StartedJobRun
[error]  required: dcos.metronome.model.JobRun => ?
[error]     runs.foreach(startJobRun)
[error]                  ^
[error] two errors found
[e

kensipe:
pushed code for the cancellable component and fixed the compiler error for the execution context for `scheduleOnce`.

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
alenkacz: yes, thank you. I like it more too...

jobs/src/test/scala/dcos/metronome/SimulatedScheduler.scala
```diff
@@ -0,0 +1,82 @@
...
+
+import akka.actor.{ Cancellable, Scheduler }
+import java.util.concurrent.atomic.AtomicLong
+
+import dcos.metronome.utils.time.FixedClock
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.FiniteDuration
+
+/**
```
alenkacz: this is copypasted from Marathon with just small alterations. I think it is not worth sharing this piece of code - e.g. in a lib or something...

alenkacz:
@kensipe ready for re-review

Important changes since you last viewed - I have added startingDeadline directly into JobRun. The main motivation was that I need to read it from somewhere upon initialization here https://github.com/dcos/metronome/pull/164/files#diff-7d090023b6b6e1f051d2252a99f01014R147

kensipe:
@alenkacz I was wondering if that would be necessary.   I will look in detail shortly

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -30,17 +32,22 @@ class JobRunExecutorActor(
...
+
+  private[impl] var startingDeadlineTimer: Option[Cancellable] = None
 
   lazy val persistenceActor = persistenceActorRefFactory(run.id, context)
   var jobRun: JobRun = run
 
   val runSpecId = jobRun.id.toPathId
 
   override def preStart(): Unit = {
+    scheduleTimeout()
```
kensipe: when we agreed to change the name to simply `scheduleTimeout` it was with the passing of startingDeadline such as `scheduleTimeout(startingDeadline)`.   now the deadline is in the job run so perhaps it is `scheduleTimeout(jobRun.startingDeadline)`.    I do not like the abbreviated name without the context of what timeout we are scheduling.

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -82,13 +89,26 @@ class JobRunExecutorActor(
     context.become(starting)
   }
 
+  def scheduleTimeout(): Unit = {
+    import scala.concurrent.duration._
+
+    jobRun.startingDeadline.foreach { deadline =>
+      val now = clock.now()
+      val from = run.createdAt.plus(deadline.toMillis)
+      val timeout = Seconds.secondsBetween(now, from).getSeconds.seconds
```
kensipe: this timeout could be nonpositive...  imagine a task that was scheduled, metronome goes now and comes up after the deadline.   
For timeout <= 0 
1. we do not want to schedule anything
2. we do want to call `cancelStartingDeadline`

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -82,13 +89,26 @@ class JobRunExecutorActor(
...
+      log.info(s"Timeout scheduled for ${jobRun.id} at $timeout")
```
kensipe: I much prefer more context in the log here.  we need to know what timeout... and we need to know the configuration and what was calculated.
`s"StartingDeadline timeout of $timeout created for JobRun ${jobRun.id} with createdAt of ${jobRun.createdAt} and startingDeadline of ${deadline} "`

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -162,11 +187,11 @@ class JobRunExecutorActor(
   }
 
   def becomeAborting(): Unit = {
-    log.info(s"Execution of JobRun ${jobRun.id} has been aborted")
     // kill all running tasks
     jobRun.tasks.values.filter(t => isActive(t.status)).foreach { t =>
       driverHolder.driver.foreach(_.killTask(t.id.mesosTaskId))
     }
+    log.info(s"Purging $runSpecId")
```
kensipe: s"Purging JobRun $runSpecId"

jobs/src/main/scala/dcos/metronome/repository/impl/kv/marshaller/JobRunMarshaller.scala
```diff
@@ -89,6 +93,10 @@ object JobRunConversions {
         status = proto.getStatus.toModel,
         createdAt = new DateTime(proto.getCreatedAt, DateTimeZone.UTC),
         completedAt = if (proto.hasFinishedAt) Some(new DateTime(proto.getFinishedAt, DateTimeZone.UTC)) else None,
+        startingDeadline = if (proto.hasStartingDeadlineSeconds)
+        Some(Duration(proto.getStartingDeadlineSeconds, TimeUnit.SECONDS))
+      else
+        None,
```
kensipe: I don't really like the alignment of this code.   It isn't consistent with completedAt.  It is longer (perhaps that is the reason to break to more than 1 line).    Now the `else` is hanging out there and it forces the reader to trace back to find the context.

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -82,13 +89,26 @@ class JobRunExecutorActor(
...
 
+  def scheduleTimeout(): Unit = {
+    import scala.concurrent.duration._
+
+    jobRun.startingDeadline.foreach { deadline =>
+      val now = clock.now()
+      val from = run.createdAt.plus(deadline.toMillis)
+      val timeout = Seconds.secondsBetween(now, from).getSeconds.seconds
+      log.info(s"Timeout scheduled for ${jobRun.id} at $timeout")
+      startingDeadlineTimer = Some(context.system.scheduler.scheduleOnce(timeout, self, StartTimeout))
```
meln1k: What do you think about using the scheduler from constructor?

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -30,17 +32,22 @@ class JobRunExecutorActor(
...
     clock:                      Clock,
     val behavior:               Behavior
-) extends Actor with Stash with ActorLogging with ActorBehavior {
+)(implicit scheduler: Scheduler) extends Actor with Stash with ActorLogging with ActorBehavior {
   import JobRunExecutorActor._
   import JobRunPersistenceActor._
   import TaskStates._
+  import context.dispatcher
+
+  private[impl] var startingDeadlineTimer: Option[Cancellable] = None
```
meln1k: What do you think about moving this var into the receive function parameter? less mutable state, less problems :)

Or we can use heavy guns like akka-FSM :)

jobs/src/test/scala/dcos/metronome/SimulatedScheduler.scala
meln1k: Maybe we can create a PR for akka :)

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
timcharper: Recommendation to avoid Akka FSM. It doesn't end well.

jobs/src/main/scala/dcos/metronome/repository/impl/kv/marshaller/JobRunMarshaller.scala
timcharper: I bet you need to update scalariform. I saw a bunch of these fixed when I upgraded Marathon.

jobs/src/main/scala/dcos/metronome/utils/time/Clock.scala
```diff
@@ -13,6 +13,11 @@ class SystemClock(dateTimeZone: DateTimeZone = DateTimeZone.UTC) extends Clock {
 }
 
 class FixedClock(var _now: DateTime) extends Clock {
+  private[this] var subscribers: List[() => Unit] = Nil
+  def onChange(fn: () => Unit): Unit = synchronized {
+    subscribers = fn :: subscribers
+  }
+
```
timcharper: Where is this subscribers used? My reading of the code indicates that it is unused.

jobs/src/test/scala/dcos/metronome/SimulatedScheduler.scala
```diff
@@ -0,0 +1,82 @@
...
+    override def isCancelled = scheduledTasks.contains(id)
+  }
+
+  clock.onChange { () => poll() }
+
+  private[this] def doCancel(id: Long) = synchronized { scheduledTasks -= id }
+  private[this] def poll(): Unit = synchronized {
+    val now = clock.now().getMillis
+    scheduledTasks.values.foreach { task =>
+      if (task.time <= now) task.action()
```
timcharper: sigh; I wrote this code originally. And there is a deadlock potential here if the Executor provided is a same-thread executor. I guess keep for now and I'll fix it in Marathon / copy over in the future.

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
timcharper: Yeah, this should totally use the constructor provided scheduler. Good catch. I think the test would depend on this? (How does the test pass without using the test-provided scheduler?)

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
alenkacz: good point, I'll change the name

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
alenkacz: thanks!

jobs/src/main/scala/dcos/metronome/repository/impl/kv/marshaller/JobRunMarshaller.scala
alenkacz: I'll do that in a separate PR

alenkacz: already done and merged

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -30,39 +32,51 @@ class JobRunExecutorActor(
...
-          case _ =>
-            self ! Initialized(tasks = Nil)
-        }
-
-      case JobRunStatus.Success => becomeFinishing(jobRun)
-
-      case JobRunStatus.Failed  => becomeFailing(jobRun)
+    val startingDeadline = computeStartingDeadline
+    if (startingDeadline.exists(d => d.toSeconds <= 0)) {
+      log.info(s"StartingDeadline reached for job ${jobRun.id}, aborting job run.")
```
kensipe: These log messages are super important IMO...   I would prefer 1) to be consistent and 2) they are as helpful as possible and 3) they provide a clear message as to "when" in the order of events this event occurred.

For this one:
how about `s"StartingDeadline timeout of ${startingDeadline} sec expired at PreStart for JobRun ${jobRun.id} created at ${jobRun.createdAt}"`
aborting should log it's own message
this way we know with 1 log message:
1) the deadline was negative or 0 and what value
2) when the job was created
3) what the id was

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
kensipe: I don't understand why we changed from the above log system...  I don't want to translate the logs (purging means aborting etc.).    And I don't know why we switched from the jobRun.id to the runSpecId.   Can we switch back to the older message please?

jobs/src/main/scala/dcos/metronome/utils/time/Clock.scala
kensipe: This is used in `SimulatedScheduler`

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
alenkacz: This was just a debug message, I have removed it... not sure why I have removed the original one :( Thanks for noticing...

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
alenkacz: thanks! hopefully better now

jobs/src/main/scala/dcos/metronome/jobrun/JobRunModule.scala
```diff
@@ -31,7 +32,7 @@ class JobRunModule(
     val persistenceActorFactory = (id: JobRunId, context: ActorContext) =>
       context.actorOf(JobRunPersistenceActor.props(id, jobRunRepository, behavior))
     JobRunExecutorActor.props(jobRun, promise, persistenceActorFactory,
-      launchQueue, taskTracker, driverHolder, clock, behavior)
+      launchQueue, taskTracker, driverHolder, clock, behavior)(actorSystem.scheduler)
```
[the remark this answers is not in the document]
timcharper: Oh, I wouldn't put the scheduler in second parameter list. But, this is a nit. Not something necessarily that I think would weren't a rejection :-)

jobs/src/main/scala/dcos/metronome/utils/time/Clock.scala
timcharper: How though? How does it get to `SimulatedScheduler`? It's a private[this] var so not accessible via the outside, and, there's only one reference in this class. I know it's possible I'm missing something, because I often do. So, what am I missing? :-)

timcharper: See it :) it's used on line 26 👇

kensipe:
rebase to master which had formatting updates based on our linter

jobs/src/test/scala/dcos/metronome/SimulatedScheduler.scala
timcharper: Simulated scheduler deadlock potential fixed here: https://github.com/mesosphere/marathon/pull/5879

kensipe: @timcharper I looked into resolving this and it does seem as simple as the PR you provided as reference.   I've created a JIRA to capture the tech debt:  https://jira.mesosphere.com/browse/METRONOME-209

PR #165, opened 2017-12-27
Adding endpoint to expose launch queue

kensipe:
adding endpoint to expose launch queue

Summary: 

JIRA issues: METRONOME-190


api/src/main/scala/dcos/metronome/api/v1/models/package.scala
```diff
@@ -248,4 +250,13 @@ package object models {
   }
 
   implicit lazy val MetronomeInfoWrites: Writes[MetronomeInfo] = Json.writes[MetronomeInfo]
+
+  implicit lazy val QueuedTaskInfoWrites: Writes[QueuedTaskInfo] = new Writes[QueuedTaskInfo] {
+    override def writes(o: QueuedTaskInfo): JsValue = Json.obj(
+      "app" -> o.runSpec.toString,
```
kensipe: looking to clean this up still

kensipe:
ready for review

api/src/main/scala/dcos/metronome/api/v1/models/package.scala
```diff
@@ -248,4 +249,39 @@ package object models {
   }
 
   implicit lazy val MetronomeInfoWrites: Writes[MetronomeInfo] = Json.writes[MetronomeInfo]
+
+  implicit lazy val MarathonRunSpecWrites: Writes[RunSpec] = new Writes[RunSpec] {
```
alenkacz: I don't feel good about exposing RunSpec (Marathon domain object) through Metronome API...

I would rather expose our own domain object that we have under control and that will not change in between marathon version... 

What do you think?

api/src/main/scala/dcos/metronome/api/v1/models/package.scala
```diff
@@ -248,4 +249,39 @@ package object models {
...
+      "args" -> runSpec.args,
+      "env" -> runSpec.env.toString(),
+      "constraints" -> runSpec.constraints.toString(),
+      "roles" -> runSpec.acceptedResourceRoles
+    )
+  }
+
+  implicit lazy val QueuedTaskInfoWrites: Writes[QueuedTaskInfo] = new Writes[QueuedTaskInfo] {
+    override def writes(taskInfo: QueuedTaskInfo): JsValue = Json.obj(
+      "runid" -> taskInfo.runSpec.id.toString,
```
alenkacz: runId

api/src/main/scala/dcos/metronome/api/v1/models/package.scala
```diff
@@ -248,4 +249,39 @@ package object models {
...
+      "toLaunch" -> taskInfo.tasksLeftToLaunch,
+      "backoff" -> taskInfo.backOffUntil.toDateTime,
+      "runspec" -> taskInfo.runSpec
```
alenkacz: runSpec?

api/src/main/scala/dcos/metronome/api/v1/models/package.scala
```diff
@@ -248,4 +249,39 @@ package object models {
...
+    )
+  }
+
+  def queuedTaskInfoMap(id: String, queuedTaskList: Seq[QueuedTaskInfo]): JsValue = {
+    Json.obj(
+      "jobID" -> id,
```
alenkacz: jobId

jobs/src/main/scala/dcos/metronome/queue/QueueService.scala
```diff
@@ -0,0 +1,18 @@
+package dcos.metronome.queue
+
+import mesosphere.marathon.core.launchqueue.LaunchQueue.QueuedTaskInfo
+
+/**
+  * Provides access to the underlying list of tasks in the launch queue.
+  *
+  */
+trait QueueService {
```
alenkacz: Queue sounds too generic to me - do you think it makes sense to make the name more specific?

jobs/src/main/scala/dcos/metronome/queue/QueueService.scala
```diff
@@ -0,0 +1,18 @@
...
+
+  def list(): Iterable[QueuedTaskInfo]
```
alenkacz: I would personally prefer to create our own domain objects with subset of Marathon's QueuedTaskInfo and RunSpec and bind to those on this layer.

I don't feel that well about passing internal Marathon objects throughout the whole project since those might change in the future version and I feel it might be valuable to isolate places where the changes would be necessary to a small layer that is transforming Marathon internal stuff to our own stuff...

Although I feel kind of strongly about this it is definitely up to discussion! :)

api/src/test/scala/dcos/metronome/api/WithApplicationComponents.scala
```diff
@@ -121,6 +122,8 @@ class MockApiComponents(context: Context) extends BuiltInComponentsFromContext(c
 
   lazy val assets: Assets = wire[Assets]
 
+  lazy val queueService: QueueService = QueueServiceFixture.simpleQueueService()
```
alenkacz: what is the motivation behind this change? I see it is never used ... ? I don't see much value in such fixture, maybe we can drop it ... ?

api/src/main/scala/dcos/metronome/api/v1/models/package.scala
kensipe: I don't feel like we are exposing it...  this is a json transformation.   I did call it runspec on L271.  we definitely shouldn't do that.

api/src/main/scala/dcos/metronome/api/v1/models/package.scala
kensipe: yeah... we translate so much from mesos terminology, marathon terminology to metronome.   I wish we all stuck more closely to mesos.   This would be a taskId in that case.   In this case, with metronome, there is a jobid, which is the id of the job.  and there are runids which is the id of each task or run of that job.

api/src/main/scala/dcos/metronome/api/v1/models/package.scala
kensipe: yeah...noted above... we should probably call this runs

api/src/test/scala/dcos/metronome/api/WithApplicationComponents.scala
kensipe: it was focused on me for the `wire[ApiModule]` below.  Doesn't compile without it :)

jobs/src/main/scala/dcos/metronome/queue/QueueService.scala
kensipe: as always it depends :) 

if we are targeting a future merge of metronome and marathon than creating a bunch of types that need refactoring every time marathon changes seems like creating work.   In general you and I have similar views.

api/src/main/resources/routes
```diff
@@ -3,6 +3,8 @@ GET /ping dcos.metronome.api.v1.controllers.ApplicationController.ping
 GET /info dcos.metronome.api.v1.controllers.ApplicationController.info
 GET /v1/metrics dcos.metronome.api.v1.controllers.ApplicationController.showMetrics
 
+GET /v1/deployments dcos.metronome.api.v1.controllers.QueueController.queue
```
kensipe: switch this to `/v1/queue`

alenkacz: good point, because for jobs there is no notion of "deployment" anyway...

kensipe:
I've spent the entire morning looking at creating a model or reusing a model to remove marathon.   It doesn't seem like an easy goal and I don't think it gets us anything.   The marathon types already exist in many places... the JobSpec is an extension of Marathon RunSpec.   The JobRun has tasks which have the marathon Task.Id.    If we agree that we want to create a model to abstract and encapsulate marathon (which I agree with)... it needs to be a separate effort.   

There are models that are close... but require the creation of a bunch of model classes which spiral into frustration.    We kind of want the JobRun without the overhead of the JobSpec... however we do want elements of the JobSpec, the RunSpec and the JobRunSpec... without the overhead of the schedule specs...

api/src/main/resources/public/api/v1/queue.raml
```diff
@@ -0,0 +1,10 @@
+get:
+  description:
+    Get the list of all jobs in the launch queue
+  is: [ secured ]
+  responses:
+    200:
+      description: "The list of jobs in the launch queue."
```
alenkacz: I think we can lose the " here

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -85,6 +85,8 @@ class JobRunExecutorActor(
   def addTaskToLaunchQueue(): Unit = {
     log.info("addTaskToLaunchQueue")
     import dcos.metronome.utils.glue.MarathonImplicits._
+    println("*** Launch ***")
```
alenkacz: I think this was commited by mistake

api/src/main/resources/public/api/v1/examples/queue_job_list.json
```diff
@@ -0,0 +1,48 @@
...
+        "id":"/fatjob2/201712292252589GUpy",
+        "roles":[
+          "*"
+        ],
+        "runSpec":{
+          "artifacts":[
+
+          ],
+          "cmd":"nuke --dry --master local",
+          "cpus":10,
```
alenkacz: I am not sure this is right. Maybe we should take the example out of some request made against metronome? Because e.g. here is cpus but we serialize it as cpu, I don't see disk being serialized, also I see docker instead of container

I am talking about this serialization code https://github.com/dcos/metronome/pull/165/files#diff-2cf575f7407491961e9d5fa3e819122eR255

api/src/test/scala/dcos/metronome/api/WithApplicationComponents.scala
```diff
@@ -121,6 +122,8 @@ class MockApiComponents(context: Context) extends BuiltInComponentsFromContext(c
 
   lazy val assets: Assets = wire[Assets]
 
+  lazy val queueService: LaunchQueueService = QueueServiceFixture.simpleQueueService()
```
alenkacz: I have removed this locally and code still compiles. Since it is not used anywhere, I would drop it...

jobs/src/test/scala/dcos/metronome/queue/QueueServiceFixture.scala
```diff
@@ -0,0 +1,11 @@
+package dcos.metronome.queue
+
+import dcos.metronome.model.QueuedJobRunInfo
+import dcos.metronome.queue.impl.LaunchQueueServiceImpl
+
+object QueueServiceFixture {
```
alenkacz: we can drop this too when we drop WithApplicationComponents.queueService

jobs/src/main/scala/dcos/metronome/queue/impl/LaunchQueueServiceImpl.scala
```diff
@@ -0,0 +1,17 @@
+package dcos.metronome.queue.impl
+
+import dcos.metronome.jobrun.impl.QueuedJobRunConverter.QueuedTaskInfoToQueuedJobRunInfo
+import dcos.metronome.model.QueuedJobRunInfo
+import dcos.metronome.queue.LaunchQueueService
+import mesosphere.marathon.core.launchqueue.LaunchQueue
+
+class LaunchQueueServiceImpl(launchQueue: LaunchQueue) extends LaunchQueueService {
```
alenkacz: I am not a big fan of "Impl" ... maybe we can go with `MarathonLaunchQueueService` ?

jobs/src/main/scala/dcos/metronome/queue/impl/LaunchQueueServiceImpl.scala
```diff
@@ -0,0 +1,17 @@
...
+
+  override def list(): scala.collection.immutable.Seq[QueuedJobRunInfo] = {
+    launchQueue.list.map(_.toModel)
+  }
+
+  override def listGroupByJobId(): scala.collection.immutable.Map[String, scala.collection.immutable.Seq[QueuedJobRunInfo]] = {
+    this.list().groupBy(_.jobid)
```
alenkacz: you can lose "this" here

api/src/main/scala/dcos/metronome/api/v1/controllers/LaunchQueueController.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import dcos.metronome.api.{ Authorization, RestController }
+import dcos.metronome.queue.LaunchQueueService
+import dcos.metronome.api.v1.models.QueuedJobRunMapWrites
+import play.api.mvc.Action
+
+class LaunchQueueController(queueService: LaunchQueueService) extends RestController {
+
+  def queue() = Action {
+    Ok(QueuedJobRunMapWrites.writes(queueService.listGroupByJobId()))
```
alenkacz: let's do `.groupBy(_.jobid)` here as we do in LaunchQueueService and then we can drop the listGroupByJobId altogether...

jobs/src/main/scala/dcos/metronome/queue/impl/LaunchQueueServiceImpl.scala
alenkacz: but if we get rid of this method altogether it does not matter

jobs/src/main/scala/dcos/metronome/jobrun/impl/QueuedJobRunConverter.scala
```diff
@@ -0,0 +1,82 @@
...
+      }
+      ConstraintSpec(
+        attribute = constraint.getField,
+        operator = Operator.names(operator.toString),
+        value = value
+      )
+    }.toList
+  }
+
+  implicit class MarathonCotainerToDockerSpec(val container: Option[Container]) extends AnyVal {
```
alenkacz: cotainer -> container

alenkacz:
@kensipe and FYI - I am already working on the controllers tests

api/src/main/resources/public/api/v1/examples/queue_job_list.json
kensipe: it was "cpu' not "cpus" which was incorrect... fixing.     This is also in RunSpec (which is the wrong place to be looking:).    This json is serialized by the JobRunSpec which serializes cpus and disk.       docker is also correct... metronome was built for docker and translates that to container.docker for marathon but has no concept of container at this point.
This is correct json

api/src/main/resources/public/api/v1/queue.raml
kensipe: we have quotes in other another raml as well... I will remove it in all ramls for consistency

jobs/src/main/scala/dcos/metronome/queue/impl/LaunchQueueServiceImpl.scala
kensipe: I would rather stay consistent with the existing code base for now which already uses "Impl"

api/src/main/scala/dcos/metronome/api/v1/controllers/LaunchQueueController.scala
kensipe: I debated this... but I like it!   Thanks!

meln1k:
Aren't we missing some tests for this new fancy endpoint? @kensipe

kensipe:
@meln1k yep.   I thought alenkacz indicated she was working on them...   I'll be looking at that shortly

timcharper:
Is it not sufficient just to show the job ID (as opposed to the full runspec)? I think I might be conservative and add less information here, since it's easier to add than take away.

api/src/main/resources/public/api/v1/examples/queue_job_list.json
```diff
@@ -0,0 +1,48 @@
+[
+  {
+    "job":"fatjob2",
```
alenkacz: shouldn't this be "id" not "job"? or "jobId"...

We use the word id for job id on most places as I have seen

kensipe:
Regarding the information return from /1/queue and including the full runspec.   I support reducing it... however I wanted information that might help understand why it is still in the queue.   That said... things that make a lot of sense to me include:   `job_run_id`, `job_id` (for id purposes) and `role(s)`, `cpu`, `mem`, `disk`, `user`, `placement`.      I also thought it made sense to just maintain the structure of job and run... what are thoughts here?   I definitely was after grouping job_runs with jobs which is debatable.

Now that the work is done... is there a push to reduce this?   thoughts?

kensipe:
I suppose if we provide job_id and run_id that the user and use that information to get the additional information... hmm.

kensipe:
on marathon `v1/deployments` looks like

```
[{"id":"06f0982d-3ad8-4255-960d-6c8f70249948","version":"2018-01-03T21:26:09.569Z","affectedApps":["/sleep"],"affectedPods":[],"steps":[{"actions":[{"action":"StartApplication","app":"/sleep"}]},{"actions":[{"action":"ScaleApplication","app":"/sleep"}]}],"currentActions":[{"action":"ScaleApplication","app":"/sleep","readinessCheckResults":[]}],"currentStep":2,"totalSteps":2}]
```

and `v1/queue` looks like:
```
{"queue":[{"count":1,"delay":{"timeLeftSeconds":0,"overdue":true},"since":"2018-01-03T21:26:10.296Z","processedOffersSummary":{"processedOffersCount":6,"unusedOffersCount":6,"lastUnusedOfferAt":"2018-01-03T21:26:20.409Z","rejectSummaryLastOffers":[{"reason":"UnfulfilledRole","declined":1,"processed":2},{"reason":"UnfulfilledConstraint","declined":0,"processed":1},{"reason":"NoCorrespondingReservationFound","declined":0,"processed":1},{"reason":"InsufficientCpus","declined":1,"processed":1},{"reason":"InsufficientMemory","declined":0,"processed":0},{"reason":"InsufficientDisk","declined":0,"processed":0},{"reason":"InsufficientGpus","declined":0,"processed":0},{"reason":"InsufficientPorts","declined":0,"processed":0}],"rejectSummaryLaunchAttempt":[{"reason":"UnfulfilledRole","declined":3,"processed":6},{"reason":"UnfulfilledConstraint","declined":0,"processed":3},{"reason":"NoCorrespondingReservationFound","declined":0,"processed":3},{"reason":"InsufficientCpus","declined":3,"processed":3},{"reason":"InsufficientMemory","declined":0,"processed":0},{"reason":"InsufficientDisk","declined":0,"processed":0},{"reason":"InsufficientGpus","declined":0,"processed":0},{"reason":"InsufficientPorts","declined":0,"processed":0}]},"app":{"id":"/sleep","backoffFactor":1.15,"backoffSeconds":1,"cmd":"sleep inf","container":{"type":"MESOS","volumes":[]},"cpus":20,"disk":0,"executor":"","instances":1,"labels":{},"maxLaunchDelaySeconds":3600,"mem":128,"gpus":0,"networks":[{"mode":"host"}],"portDefinitions":[],"requirePorts":false,"upgradeStrategy":{"maximumOverCapacity":1,"minimumHealthCapacity":1},"version":"2018-01-03T21:26:09.569Z","versionInfo":{"lastScalingAt":"2018-01-03T21:26:09.569Z","lastConfigChangeAt":"2018-01-03T21:26:09.569Z"},"killSelection":"YOUNGEST_FIRST","unreachableStrategy":{"inactiveAfterSeconds":0,"expungeAfterSeconds":0}}}]}
```

kensipe:
we seem to be closer to the deployment endpoint than the queue endpoint.   Should we be consist?   should we call this `deployments` ?  and provide `job_id` with a list of `run_id`?

kensipe:
btw... the marathon queue does have the runspec

kensipe:
reducing output based on feedback and conversations to this:
```
[
    {
        "job": "test-deadline3",
        "runs": [
            "/test-deadline3/20180103214014JIT0Y",
            "/test-deadline3/20180103214004BPuOB"
        ]
    }
]```

**Comment by kensipe** [comment]
still need to **decide** endpt name:  `v1/queue` or `v1/deployments`

**Comment by alenkacz** [comment]
I would go with /queue since jobs are not having much for of a "deployment" concept

Also I would go with this API - it gives us more flexibility to include more runspec things in the future releases without breaking compatibility (objects instead of just array of strings).
{
       "job": "test-deadline3",
       "runs": [
           {
        “runId”: "/test-deadline3/20180103214014JIT0Y"
    },
    {
        “runId”: "/test-deadline3/20180103214004BPuOB"
    }
       ]
   }

File: jobs/src/main/scala/dcos/metronome/queue/impl/LaunchQueueServiceImpl.scala (hunk 15 above)
Comment (timcharper) [reply jobs/src/main/scala/dcos/metronome/queue/impl/LaunchQueueServiceImpl.scala]: Agreed with both

my preference is to name the traits "LaunchQueueServiceLike" and them omit from Impl.

But consistency is valuable.

File: api/src/main/scala/dcos/metronome/api/v1/models/package.scala [hunk 20]
```diff
@@ -251,4 +253,43 @@ package object models {
   }
 
   implicit lazy val MetronomeInfoWrites: Writes[MetronomeInfo] = Json.writes[MetronomeInfo]
+
+  implicit lazy val MarathonRunSpecWrites: Writes[RunSpec] = new Writes[RunSpec] {
+    override def writes(runSpec: RunSpec): JsValue = Json.obj(
```
timcharper: seems like this is dead code now?

kensipe:
@timcharper roger that... on it.   What do you think of the API change suggested by @alenkacz ?   I was looking to move to it... we would love you feedback there.

kensipe:
removed dead code and updated output to:

```
[
    {
        "jobId": "test-deadline3",
        "runs": [
            {
                "runId": "/test-deadline3/20180103214004BPuOB"
            },
            {
                "runId": "/test-deadline3/20180103214014JIT0Y"
            }
        ]
    }
]```

PR #166, opened 2017-12-30
Make Docker.forcePullImage Optional

kensipe:
Make Docker.forcePullImage Optional

Summary:
making forcePullImage optional.  I thought Big B boolean would do it.  Apparently we need the Optional.
https://jira.mesosphere.com/browse/METRONOME-196

JIRA issues:  METRONOME-196

### DIFF
--- a/jobs/src/main/scala/dcos/metronome/model/Container.scala
+++ b/jobs/src/main/scala/dcos/metronome/model/Container.scala
@@ -2,5 +2,5 @@ package dcos.metronome.model
 
 trait Container
 
-case class DockerSpec(image: String, forcePullImage: Boolean = false) extends Container
+case class DockerSpec(image: String, forcePullImage: Option[Boolean] = Some(false)) extends Container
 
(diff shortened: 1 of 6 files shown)


jobs/src/main/scala/dcos/metronome/model/Container.scala
```diff
@@ -2,5 +2,5 @@ package dcos.metronome.model
 
 trait Container
 
-case class DockerSpec(image: String, forcePullImage: Boolean = false) extends Container
+case class DockerSpec(image: String, forcePullImage: Option[Boolean] = Some(false)) extends Container
```
alenkacz: I would say boolean with a default value should be enough... why not? I would like to see a failing test to see why this is necessary... I tried looking for such validator but did not find any...

What exactly is failing?

kensipe: off master try creating a job with a docker (include a docker image and no forcePullImage).  It will fail.  It currently fails on the version of metronome released to 1.11 testing/master.

To be fair... I am not that familiar with the validation framework we are using.  I thought that the "required" in the JSON schema would be respected.  image is required, forcePullImage is not.   That is not the case.   What is provided here provides a working solution.

kensipe:
Better solution with #167

PR #158, opened 2017-12-13
Adding Force Pull Image

kensipe:
Adding Docker image force pull according to  https://jira.mesosphere.com/browse/METRONOME-148

This is based on the changes made by graceful task killing which needs to land first. 

JIRA:  METRONOME-148

PR #172, opened 2018-01-09
Support Docker forcePullImage for v0

kensipe:
Support Docker forcePullImage for v0

Summary:
https://jira.mesosphere.com/browse/METRONOME-207

JIRA issues:  METRONOME-207

### DIFF
--- a/api/src/main/resources/public/api/v0/schema/jobspec_v0.schema.json
+++ b/api/src/main/resources/public/api/v0/schema/jobspec_v0.schema.json
@@ -89,7 +89,7 @@
             },
             "forcePullImage": {
               "type": "boolean",
-              "documentation": "The containerizer will pull the image from the registry, even if the image is already downloaded on the agent."
+              "documentation": "The containerizer will pull the image from the registry, even if the image is already downloaded on the agent node."
             }
           },
           "required": ["image"]
--- a/api/src/main/resources/public/api/v0/schema/jobspec_v0.schema.json
+++ b/api/src/main/resources/public/api/v0/schema/jobspec_v0.schema.json
@@ -86,6 +86,10 @@
             "image": {
               "type": "string",
               "documentation": "The docker repository image name."
+            },
+            "forcePullImage": {
+              "type": "boolean",
+              "documentation": "The containerizer will pull the image from the registry, even if the image is already downloaded on the agent."
             }
           },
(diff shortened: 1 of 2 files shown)


api/src/main/resources/public/api/v0/schema/jobspec_v0.schema.json
```diff
@@ -86,6 +86,10 @@
             "image": {
               "type": "string",
               "documentation": "The docker repository image name."
+            },
+            "forcePullImage": {
+              "type": "boolean",
+              "documentation": "The containerizer will pull the image from the registry, even if the image is already downloaded on the agent."
```
ichernetsky: `on the agent` -> ` on the agent node`?

api/src/test/scala/dcos/metronome/api/v1/controllers/JobSpecControllerTest.scala
```diff
@@ -45,6 +45,30 @@ class JobSpecControllerTest extends PlaySpec with OneAppPerTestWithComponents[Mo
       contentType(response) mustBe Some("application/json")
     }
 
+    "creates job when sending docker spec with forcePullImage property against v0" in {
+      Given("Job spec with forcePullImage")
+      val specJson = "{\"id\":\"spec1\",\"run\":{\"cpus\":1,\"mem\":128,\"disk\":0,\"docker\":{\"image\":\"image\", \"forcePullImage\": true}}}"
```
ichernetsky: Maybe it is worth using triple quotes here? Like
```
@ """{"foo":"bar"}""" 
res0: String = "{\"foo\":\"bar\"}"
```

kensipe: I was stay consistent... and truth be known, I copied a previous test and modified for this purpose... but yeah... this should be the way

timcharper:
How does this signal get ultimately passed to Mesos?

timcharper:
Was it already done but v0 api just needed JSON schema change?

kensipe:
most of the work was in #158 however this required forcepullimage which was fixed by #167 however all of this work was on the `/v1` endpoint and not `/v0`.   It turns out the UI uses the v0 end point.

kensipe:
rebased

PR #174, opened 2018-01-10
Fix for Metronome Restart Starting Additional Jobs

kensipe:
Fix for Metronome Restart Starting Additional Jobs

Summary:
Metronome if restarted while a JobRun was running upon startup would launch a new Task for the existing JobRun.  This would create a situation where there are tasks that it appeared Metronome wasn't aware of and it would put heavier demands on the cluster.
This initially looked like a reconciliation issue, however the reconciliation of known tasks looks fine with this fix.   There is still an issue IMO in the odd state where mesos think there is a metronome task but metronome has no record of it.  That will need to be address in another JIRA. 
A contributing factor is that it is expected that the launch queue contains all the launched tasks.  However that is not true after a restart.  All previously launched and running "active" tasks are not in the launch queue.

This PR resolves the issue of metronome starting another task on restart which resulted in 1 jobrun having multiple tasks associated with it which should never happen.   We allow multiple jobruns for a job but not multiple tasks for a jobrun.   It is always expected a 1 to 1 relationship for jobrun and task.    Another side effect of this bug is metronome would lose track of a task... this was a 2nd task for a jobrun.  When the 1st task completed it would be purged (besides history) from metronome which is the correct behavior however the 2nd (or more depending on the number of restarts) task would be running on mesos without metronome awareness.  This is resolved with this fix.

https://jira.mesosphere.com/browse/METRONOME-100
https://jira.mesosphere.com/browse/METRONOME-139
https://jira.mesosphere.com/browse/METRONOME-189

JIRA issues: METRONOME-100

### DIFF
--- a/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
@@ -46,36 +46,25 @@ class JobRunExecutorActor(
   val runSpecId = jobRun.id.toPathId
 
   override def preStart(): Unit = {
-    val startingDeadline = computeStartingDeadline
-    if (startingDeadline.exists(d => d.toSeconds <= 0)) {
-      log.info(s"StartingDeadline timeout of ${jobRun.startingDeadline.get} expired for JobRun ${jobRun.id} created at ${jobRun.createdAt}")
-      becomeAborting()
-    } else {
-      startingDeadline.foreach(scheduleStartingDeadline)
+    // only 2 ways we should be here... 1) newly launched jobrun and 2) startup while last known jobruns were launched
+    log.info(s"Starting jobrun ${jobRun} with id ${jobRun.id} status: ${jobRun.status}")
 
-      jobRun.status match {
-        case JobRunStatus.Initial => becomeCreating()
+    jobRun.status match {
+      // on new jobrun
+      case JobRunStatus.Initial => becomeCreating()
 
-        case JobRunStatus.Starting | JobRunStatus.Active =>
-          launchQueue.get(runSpecId) match {
-            case Some(info) if info.finalTaskCount > 0 =>
-              log.info("found an active launch queue for {}, not scheduling more tasks", runSpecId)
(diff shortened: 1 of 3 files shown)


kensipe:
The main solution is here and tested locally.   Tests are broken and need to be fixed before we are ready for review.

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -46,36 +42,25 @@ class JobRunExecutorActor(
...
+    log.info(s"Starting jobrun ${jobRun} with id ${jobRun.id} status: ${jobRun.status}")
+
+    jobRun.status match {
+      // on new jobrun
+      case JobRunStatus.Initial => becomeCreating()
+
+      // previously launched job (from zk) only possible with restart
+      case JobRunStatus.Starting =>
+        println(s"Initializing Actor for existing jobrun ${jobRun} in Starting state not in the queue.")
+        context.become(creating)
```
alenkacz: I don't think this would work... [ersistence actor will never send the expected messages so we would never move out of this state

alenkacz: maybe you meant starting...

kensipe: nope...   The secret here and the reason it works is this is `preStart` which initializes this actor.     This is called for 2 reasons that I can think of :)
1. We are starting a jobrun (which has never been started before).  It will come into `preStart` with the Initial and become creating
2. we are restarting metronome where jobruns were previously started.   This actor is initialized with a jobrun in an undefined status (it is whatever the status is at the time metronome died.   In the case of Initial everything is as normal.   In the case of Starting,  things are different.   1 of 2 things are true... the task was never launched in which case we need to create it again... or the task was launched but we died before the status update to active occurred.   Until I wrote this response I had not considered the latter case but was coded for the former case.

Once we leave `preStart`.    lifecycle is managed by receive.
now I need to think about the other part of Starting case...

kensipe:
Work is superseded by #177

PR #176, opened 2018-01-11
/v1/queue should only return jobruns that are deploying

kensipe:
/v1/queue should only return jobruns that are deploying

Summary:
queue request should only return tasks inprogress not tasks that actively running

JIRA issues: METRONOME-190


kensipe:
I would like to get the current implementation of queue fixed which this PR does.    Lets add auth as a separate PR

kensipe:
and... to your question.... I definitely agree

PR #177, opened 2018-01-12
New ExecutorActor flow without launching extra tasks

alenkacz:
Review first by reading the original implementation to see the weaknesses, then proceed to the new implementation.

The problem we saw in the old implementation: when metronome process was shutdown and job run was already active or running, launch queue was called https://github.com/dcos/metronome/compare/av/reconciliation?expand=1#diff-b17aaa408214f692ce5d2015f5d1945fL60 but for a freshly started process, that does not contain anything so instead metronome launched a new job run.

The new implementation changes this flow to the following:
- ask task tracker if has known tasks, if so, use those to initialize actor
- if not known to task tracker, they might already be in the queue
- if not in the queue, proceed to launching a new task

JIRA issues: METRONOME-100

### DIFF
--- a/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
@@ -6,7 +6,7 @@ import dcos.metronome.{ JobRunFailed, UnexpectedTaskState }
 import dcos.metronome.behavior.{ ActorBehavior, Behavior }
 import dcos.metronome.eventbus.TaskStateChangedEvent
 import dcos.metronome.jobrun.StartedJobRun
-import dcos.metronome.model.{ JobResult, JobRun, JobRunId, JobRunStatus, JobRunTask, RestartPolicy }
+import dcos.metronome.model._
 import dcos.metronome.scheduler.TaskState
 import dcos.metronome.utils.time.Clock
 import mesosphere.marathon.MarathonSchedulerDriverHolder
@@ -32,13 +32,15 @@ class JobRunExecutorActor(
   taskTracker:                TaskTracker,
   driverHolder:               MarathonSchedulerDriverHolder,
   clock:                      Clock,
-  val behavior:               Behavior)(implicit scheduler: Scheduler) extends Actor with Stash with ActorLogging with ActorBehavior {
+  val behavior:               Behavior,
+  reconciliationTimeout:      FiniteDuration)(implicit scheduler: Scheduler) extends Actor with Stash with ActorLogging with ActorBehavior {
   import JobRunExecutorActor._
   import JobRunPersistenceActor._
   import TaskStates._
   import context.dispatcher
 
   private[impl] var startingDeadlineTimer: Option[Cancellable] = None
+  private[impl] var reconciliationTimer: Option[Cancellable] = None
 
   lazy val persistenceActor = persistenceActorRefFactory(run.id, context)
   var jobRun: JobRun = run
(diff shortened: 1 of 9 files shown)


jobs/src/main/scala/dcos/metronome/scheduler/impl/ReconciliationActor.scala
```diff
@@ -54,7 +56,7 @@ class ReconciliationActor(
 
   onTransition {
     case _ -> Idle =>
-      // TODO: eventBus.publish(ReconciliationFinished)
+      context.system.eventStream.publish(ReconciliationFinished())
```
alenkacz: I know this is not used anymore but I guess there is no harm in keeping it already in place, right?

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -47,35 +48,31 @@ class JobRunExecutorActor(
...
+        becomeAborting()
+      case JobRunStatus.Initial =>
+        becomeCreating()
+      case JobRunStatus.Starting | JobRunStatus.Active =>
+        val existingTasks = tasksFromTaskTracker()
+        existingTasks match {
+          case tasks if tasks.nonEmpty =>
+            // the actor was probably just restarted and we did not lose contents of the launch queue
+            self ! Initialized(tasks)
+          case _ if jobRun.status == JobRunStatus.Starting && existsInLaunchQueue() =>
```
alenkacz: is my assumption correct, that the situation for item being in launch queue but not in task tracker can happen only for starting task?

jobs/src/main/scala/dcos/metronome/scheduler/impl/ReconciliationActor.scala
meln1k: looks good :)

alenkacz:
btw. once I got approvals that my direction is correct I will proceed to some test coverage :)

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -39,6 +39,7 @@ class JobRunExecutorActor(
   import context.dispatcher
 
   private[impl] var startingDeadlineTimer: Option[Cancellable] = None
+  private[impl] var reconciliationTimer: Option[Cancellable] = None
```
meichstedt: this seems unused now

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
meichstedt: Yes

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -211,6 +220,10 @@ class JobRunExecutorActor(
     context.become(aborting)
   }
 
+  def proceedAfterReconciliation(): Unit = {
```
meichstedt: This is also unused, right?

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunServiceActor.scala
```diff
@@ -29,6 +29,7 @@ class JobRunServiceActor(
   override def preStart(): Unit = {
     super.preStart()
     context.system.eventStream.subscribe(self, classOf[TaskStateChangedEvent])
+    context.system.eventStream.subscribe(self, classOf[Event.ReconciliationFinished])
```
meichstedt: unused

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -131,6 +127,17 @@ class JobRunExecutorActor(
     startingDeadlineTimer = None
   }
 
+  def tasksFromTaskTracker(): Iterable[JobRunTask] = {
+    taskTracker.appTasksLaunchedSync(runSpecId).collect {
+      // FIXME: we do currently only allow non-resident tasks. Since there is no clear state on
```
kensipe: lets remove this FIXME.   I'm fine with noting that we only work with Ephemeral...   Any change to this should be represented in a JIRA not a FIXME note.

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -371,6 +379,7 @@ object JobRunExecutorActor {
   case class Aborted(jobResult: JobResult)
 
   case object StartTimeout
+  case object ReconciliationTimeout
```
kensipe: I do NOT find a reference for this

PR #179, opened 2018-01-16
Introduce SI test for metronome restart

alenkacz:
Summary:
Disclaimer: Right now this should fail on currently released Metronome and it does. When METRONOME-100 if fixed, this will be green again.

We had a bug:
- start a job run
- restart metronome before the job run finishes
- on start-up, metronome ignores existing mesos task and spins up new task for that job run

JIRA issues: METRONOME-100


tests/system/test_root_metronome.py
```diff
@@ -239,6 +239,28 @@ def test_docker_job():
...
+def test_metronome_shutdown_with_no_extra_tasks():
+    """ Test for METRONOME-100 regression
+        When Metronome is restarted it incorrectly started another task for already running job run task.
+    """
+    client = metronome.create_client()
+    job_id = "metronome-shutdown-{}".format(uuid.uuid4().hex)
+    with job(job_no_schedule(job_id)):
+        # run a job before we shutdown Metronome
+        client.run_job(job_id)
+        time.sleep(2)
```
alenkacz: I know that this is potentially flaky but this is a common pattern in these tests right now. I would rather change that in a separate PR addressing this https://jira.mesosphere.com/browse/METRONOME-212

tests/system/test_root_metronome.py
```diff
@@ -239,6 +239,28 @@ def test_docker_job():
...
+        common.assert_one_job_run_with_one_task(client, job_id)
+
+        # restart metronome process
+        metronome_leader = shakedown.marathon_leader_ip()
+        shakedown.run_command_on_agent(metronome_leader, 'sudo systemctl restart dcos-metronome')
+        shakedown.time_wait(lambda: common.metronome_available_predicate(), timeout_seconds=timedelta(minutes=1).total_seconds())
+
+        # verify that no extra job runs were started when Metronome was restarted
+        time.sleep(20) # launching a new task takes some time, we have to give marathon enough time to do that
```
kensipe: the way we manage this in marathon (which wasn't available in metronome until recently) is to check the deployment queue.   We should do that here for the `time.sleep(2)` and for the `time.sleep(20)`.     It is best if we add this call to the metronome class in dcos-cli... but that will take time to propagate for our use here.   What you want to do is review marathon.deployment_wait https://github.com/dcos/shakedown/blob/e711e691d9692b06137465da61f0b3c6f7a53131/shakedown/dcos/marathon.py#L70
for metronome hit /metronome/v1/queue and wait for the json return of `[]` (up until a timeout)

tests/system/test_root_metronome.py
```diff
@@ -239,6 +239,28 @@ def test_docker_job():
...
+    job_id = "metronome-shutdown-{}".format(uuid.uuid4().hex)
+    with job(job_no_schedule(job_id)):
+        # run a job before we shutdown Metronome
+        client.run_job(job_id)
+        time.sleep(2)
+        common.assert_one_job_run_with_one_task(client, job_id)
+
+        # restart metronome process
+        metronome_leader = shakedown.marathon_leader_ip()
+        shakedown.run_command_on_agent(metronome_leader, 'sudo systemctl restart dcos-metronome')
```
kensipe: it is unclear if this will provide a good test under a multi-master situation... I think it will but I haven't tested it.   I'm not sure if this condition happens at the point of becoming leader or at the point of startup (as leader).   However... in a multi-master environment... the `time_wait` below will return almost immediately without a full restart taking place as the standby metronome takes over leadership.

alenkacz: good point, I'll test it in multi-master setup as well

tests/system/test_root_metronome.py
alenkacz: I am not sure that deployment wait will solve this, because I think there might be a time, when metronome is up (API is returning HTTP 200) but executor actor did not manage to put anything into the queue...

regarding that queue we should definitely take advantage of that. I added a ticket for that yesterday

tests/system/test_root_metronome.py
alenkacz: I tested it and it works ok. I think when leader takes over the new leader have to load the state from the ZK as there is no in-memory state in the stand-by leader. But correct me if I am wrong...

Btw. it would be nice to add /leader endpoint to metronome

alenkacz: oh sorry, on multimaster the test PASSES which is weird... will have to look into this once more tomorrow

alenkacz: I added a comment - I think right now there is no easy way how to make this work in multimaster setup

tests/system/common.py
```diff
@@ -83,3 +84,22 @@ def wait_for_mesos_endpoint(timeout_sec=timedelta(minutes=5).total_seconds()):
...
+def metronome_available_predicate():
+    url = metronome_api_url()
+    try:
+        response = http.get(url)
+        return response.status_code == 200
+    except Exception as e:
+        return False
+
+def metronome_api_url():
+    return shakedown.dcos_url_path("/service/metronome/v1/jobs")
```
kensipe: `shakedown.dcos_service_url("metronome/v1/jobs")` ?  not a big deal

alenkacz: that does not work :( because what that method does is: "/service/metronome/v1/jobs/" - it puts slash after the last word which is not what I want :(

tests/system/common.py
```diff
@@ -83,3 +84,22 @@ def wait_for_mesos_endpoint(timeout_sec=timedelta(minutes=5).total_seconds()):
...
+
+
+def assert_one_job_run_with_one_task(client, job_id):
```
zen-dog: Let's make it a bit more generic. How about `def assert_job_run(client, job_id, runs_number=1, tasks_number=1)`?

tests/system/test_root_metronome.py
zen-dog: Those sleeps are meh. We tried to remove them in marathon tests - no need to start with this here. Parsing the result of `run_job(job_id)` should give you the jobId (e.g. `"id": "20160715130259A34HX"`). Querying that from `v1/jobs& activeRuns=true&history=true&scheduled=true` should show when the run ends. So `deployment_wait`.

tests/system/test_root_metronome.py
zen-dog: This is one possibility. Querying `v1/jobs` like I posted above is another.

tests/system/test_root_metronome.py
```diff
@@ -239,6 +239,30 @@ def test_docker_job():
...
+    with job(job_no_schedule(job_id)):
+        # run a job before we shutdown Metronome
+        client.run_job(job_id)
+        time.sleep(2)
+        common.assert_one_job_run_with_one_task(client, job_id)
+
+        # restart metronome process
+        # this won't work in multi-master setup if the mesos leader is not the same as metronome leader
+        # we can improve this one there is a good way how to get metronome leader from the system (e.g. info endpoint)
+        metronome_leader = shakedown.master_leader_ip()
```
zen-dog: Right now you can mark this test with `@shakedown.masters(1)` to run only on a single-master cluster

tests/system/common.py
```diff
@@ -83,3 +85,43 @@ def wait_for_mesos_endpoint(timeout_sec=timedelta(minutes=5).total_seconds()):
...
+    shakedown.time_wait(lambda: job_run_predicate(job_id, run_id),
+              timeout)
+
+
+def assert_wait_for_no_additional_tasks(client, job_id, timeout=20, tasks_count=1):
+    """ Starting Metronome and all its actors takes some time and there is no way how to query API to figure out it finished.
+        Here we wait for given time and then assert that expected tasks count matches.
+        This covers a bug regression of METRONOME-100
+    """
+    time.sleep(timeout)
```
alenkacz: I did not figure out a way hot to get rid of this sleep part.

I need to wait for some actors to be spawned up after restart and I am not sure there is a way hot to figure out from API that it already happened...

zen-dog: So we wait to see if metronome decides to run the job? I guess we could retry for a while and fail if we find a new `scheduled`, `active` or `history` run. But that's basically the same.

tests/system/common.py
```diff
@@ -83,3 +85,43 @@ def wait_for_mesos_endpoint(timeout_sec=timedelta(minutes=5).total_seconds()):
...
+    except Exception as e:
+        return False
+
+
+def metronome_api_url():
+    return shakedown.dcos_url_path("/service/metronome/v1/jobs")
+
+
+def assert_job_run(client, job_id, runs_number=1, tasks_number=1):
+    job_runs_count = len(client.get_runs(job_id))
```
zen-dog: Now that I'm thinking about it - does `get_runs` give you active runs? scheduled? past? All 3 of them?

tests/system/common.py
alenkacz: yeah exactly ... I actually rewrote it in the way you are proposing but then I actually realized it is more or less the same as fixed wait for that time and then assert...

tests/system/common.py
alenkacz: active

zen-dog: Shouldn't we test for all 3 of them? In case the job run already finished (or scheduler?). Does it make sense?

alenkacz: yes, makes sense, I'll fix it

alenkacz: I found out there is probably no way how to get finished runs https://dcos.github.io/metronome/docs/generated/api.html# this is the closest thing and I am not sure it's worth it https://avarkocko-elasticl-c24kttf2fnwz-1817360483.us-west-2.elb.amazonaws.com/service/metronome/v1/jobs/test?embed=history

tests/system/common.py
```diff
@@ -94,12 +95,33 @@ def metronome_available_predicate():
...
+
 def metronome_api_url():
     return shakedown.dcos_url_path("/service/metronome/v1/jobs")
 
 
-def assert_one_job_run_with_one_task(client, job_id):
+def assert_job_run(client, job_id, runs_number=1, tasks_number=1):
     job_runs_count = len(client.get_runs(job_id))
-    assert job_runs_count == 1, "Expecting 1 job run but found {} for job {}.".format(job_runs_count, job_id)
+    assert job_runs_count == runs_number, "Expecting 1 job run but found {} for job {}.".format(job_runs_count, job_id)
```
zen-dog: `s/Expecting 1/Expecting ${runs_number}`

tests/system/common.py
```diff
@@ -94,12 +95,33 @@ def metronome_available_predicate():
...
     job_run_tasks = client.get_runs(job_id)[0]["tasks"]
-    assert len(job_run_tasks) == 1, "Expecting 1 job run task but found {} for job {}: {}.".format(len(job_run_tasks), job_id, job_run_tasks)
+    assert len(job_run_tasks) == tasks_number, "Expecting 1 job run task but found {} for job {}: {}.".format(len(job_run_tasks), job_id, job_run_tasks)
```
zen-dog: same here

tests/system/common.py
```diff
@@ -83,3 +85,43 @@ def wait_for_mesos_endpoint(timeout_sec=timedelta(minutes=5).total_seconds()):
...
+        return False
+
+
+def metronome_api_url():
+    return shakedown.dcos_url_path("/service/metronome/v1/jobs")
+
+
+def assert_job_run(client, job_id, runs_number=1, tasks_number=1):
+    job_runs_count = len(client.get_runs(job_id))
+    assert job_runs_count == runs_number, "Expecting 1 job run but found {} for job {}.".format(job_runs_count, job_id)
```
zen-dog: Not sure why my previous comment was marked as outdated. Anyway `s/Expecting 1/Expecting ${runs_number}`

tests/system/common.py
```diff
@@ -83,3 +85,43 @@ def wait_for_mesos_endpoint(timeout_sec=timedelta(minutes=5).total_seconds()):
...
+    job_run_tasks = client.get_runs(job_id)[0]["tasks"]
+    assert len(job_run_tasks) == tasks_number, "Expecting 1 job run task but found {} for job {}: {}.".format(len(job_run_tasks), job_id, job_run_tasks)
```
zen-dog: and the same here

PR #180, opened 2018-01-18
Queue job runs should show just job run id

kensipe:
Queue job runs should show just job run id

Summary:
current returned id exposes the marathon app id... we should just return what is known as the job runid.

JIRA issues:  METRONOME-190

### DIFF
--- a/api/src/main/scala/dcos/metronome/api/v1/models/package.scala
+++ b/api/src/main/scala/dcos/metronome/api/v1/models/package.scala
@@ -249,7 +249,7 @@ package object models {
   implicit lazy val QueuedTaskInfoWrites: Writes[QueuedJobRunInfo] = new Writes[QueuedJobRunInfo] {
     //    output of queue is only runid
     override def writes(runInfo: QueuedJobRunInfo): JsValue = Json.obj(
-      "runId" -> JsString(runInfo.id.path.last))
+      "runId" -> JsString(runInfo.runId))
   }
 
   def queuedTaskInfoMap(job: String, queuedTaskList: Seq[QueuedJobRunInfo]): JsValue = {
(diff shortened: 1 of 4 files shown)


api/src/main/scala/dcos/metronome/api/v1/models/package.scala
```diff
@@ -249,7 +249,7 @@ package object models {
   implicit lazy val QueuedTaskInfoWrites: Writes[QueuedJobRunInfo] = new Writes[QueuedJobRunInfo] {
     //    output of queue is only runid
     override def writes(runInfo: QueuedJobRunInfo): JsValue = Json.obj(
-      "runId" -> JsString(runInfo.id.toString()))
```
alenkacz: I think that we have to return jobId and runId then (as two fields). WDYT?

Because for the info to be useful for the users, they need enough informations to get more info from this endpoint
/v1/jobs/{jobId}/runs/{runId}

kensipe: we already do that... the return is { jobid: "xyz, runs: [{ runid: "test}]}

alenkacz: oh, you're absolutely right! then please also change queue_job_list.json or other places where the old thing can be

alenkacz: also please create a lazy val runId on QueuedJobRunInfo that would do this `runInfo.id.path.last` and then just call this val from here

kensipe: yep yep... good eyes and thought!

PR #181, opened 2018-01-18
Support FORBID Concurrency Policy

kensipe:
Support FORBID Concurrency Policy

Summary:
Support FORBID concurrency policy, which is to not allow the start of job run based on a schedule 
if there is currently a running job run for the job.
https://jira.mesosphere.com/browse/METRONOME-194

JIRA issues: METRONOME-194

### DIFF
--- a/jobs/src/main/scala/dcos/metronome/jobspec/impl/JobSpecSchedulerActor.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobspec/impl/JobSpecSchedulerActor.scala
@@ -4,10 +4,11 @@ package jobspec.impl
 import akka.actor._
 import dcos.metronome.behavior.{ ActorBehavior, Behavior }
 import dcos.metronome.jobrun.JobRunService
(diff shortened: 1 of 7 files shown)


jobs/src/main/scala/dcos/metronome/jobspec/impl/JobSpecSchedulerActor.scala
```diff
@@ -49,10 +50,27 @@ class JobSpecSchedulerActor(
...
+      runService.startJobRun(spec, Some(schedule.startingDeadline))
+    } else {
+      log.debug(s"Skipping Scheduled run for ${spec.id} based on concurrency policy")
+    }
     scheduleNextRun()
   }
 
+  def activeRunsForJobSpec(): Boolean = {
+    try {
+      val runResult = Await.result(runService.activeRuns(spec.id), 5.seconds)
```
meln1k: This is a bit dangerous to do blocking calls inside the receive method of the actor. How about piping the future back to the actor?

kensipe:
Going with Option 2 in #182

PR #182, opened 2018-01-18
Support FORBID Concurrency Policy (Option 2)

kensipe:
Support FORBID Concurrency Policy (Option 2)

Summary:
Support FORBID concurrency policy, which is to not allow the start of job run based on a schedule 
if there is currently a running job run for the job.
https://jira.mesosphere.com/browse/METRONOME-194

#181 has blocking code in an actor awaiting another actor.   This version moves the logic to that dependent (and context aware) actor.   This removes the block, but puts some schedule logic in the `JobRunServiceActor` which mainly used for a view on all running job runs.

If we stick with this approach, I may be inclined to have the `schedule: Option[ScheduleSpec]` hold the value of the ` startingDeadline: Option[Duration]` but I'm not sure if that makes sense in all cases.

JIRA issues: METRONOME-194

### DIFF
--- a/jobs/src/main/scala/dcos/metronome/jobrun/JobRunService.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobrun/JobRunService.scala
@@ -1,7 +1,7 @@
 package dcos.metronome
 package jobrun
 
-import dcos.metronome.model.{ JobId, JobRun, JobRunId, JobSpec }
+import dcos.metronome.model._
 
 import scala.concurrent.Future
 import scala.concurrent.duration.Duration
@@ -41,7 +41,7 @@ trait JobRunService {
     * @param jobSpec the specification to run.
     * @return the started job run
     */
-  def startJobRun(jobSpec: JobSpec, startingDeadline: Option[Duration] = None): Future[StartedJobRun]
+  def startJobRun(jobSpec: JobSpec, schedule: Option[ScheduleSpec] = None, startingDeadline: Option[Duration] = None): Future[StartedJobRun]
 
   /**
     * Kill a given job run by the given job run id.
--- a/jobs/src/main/scala/dcos/metronome/jobrun/JobRunService.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobrun/JobRunService.scala
@@ -41,7 +41,7 @@ trait JobRunService {
     * @param jobSpec the specification to run.
     * @return the started job run
     */
-  def startJobRun(jobSpec: JobSpec, schedule: Option[ScheduleSpec] = None, startingDeadline: Option[Duration] = None): Future[StartedJobRun]
+  def startJobRun(jobSpec: JobSpec, schedule: Option[ScheduleSpec] = None): Future[StartedJobRun]
 
   /**
     * Kill a given job run by the given job run id.
--- a/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunServiceActor.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunServiceActor.scala
@@ -64,12 +64,9 @@ class JobRunServiceActor(
   def triggerJobRun(spec: JobSpec, schedule: Option[ScheduleSpec], startingDeadline: Option[Duration]): Unit = {
     log.info(s"Trigger new JobRun for JobSpec: $spec")
 
-    val skipRun = schedule match {
-      case Some(schedule) => schedule.concurrencyPolicy == ConcurrencyPolicy.Forbid && runsForJob(spec.id).nonEmpty
-      case None           => false
-    }
+    val skipRun = schedule.exists(schedule => schedule.concurrencyPolicy == ConcurrencyPolicy.Forbid && runsForJob(spec.id).nonEmpty)
     if (skipRun) {
-      log.debug(s"SkippingB Scheduled run for ${spec.id} based on concurrency policy")
+      log.debug(s"Skipping scheduled run for ${spec.id} based on concurrency policy")
     } else {
       val jobRun = JobRun(JobRunId(spec), spec, JobRunStatus.Initial, clock.now(), None, startingDeadline, Map.empty)
       val startedJobRun = startJobRun(jobRun)
--- a/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunServiceActor.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunServiceActor.scala
@@ -66,12 +66,9 @@ class JobRunServiceActor(
 
     val skipRun = schedule.exists(schedule => schedule.concurrencyPolicy == ConcurrencyPolicy.Forbid && runsForJob(spec.id).nonEmpty)
     if (skipRun) {
-      log.debug(s"Skipping scheduled run for ${spec.id} based on concurrency policy")
+      log.info(s"Skipping scheduled run for ${spec.id} based on concurrency policy")
     } else {
-      val startingDeadline: Option[Duration] = schedule match {
-        case Some(schedule) => Some(schedule.startingDeadline)
-        case _              => None
-      }
+      val startingDeadline: Option[Duration] = schedule.map(_.startingDeadline)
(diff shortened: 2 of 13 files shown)


jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunServiceActor.scala
```diff
@@ -42,31 +41,40 @@ class JobRunServiceActor(
...
-    val jobRun = JobRun(JobRunId(spec), spec, JobRunStatus.Initial, clock.now(), None, startingDeadline, Map.empty)
-    val startedJobRun = startJobRun(jobRun)
-    sender() ! startedJobRun
+
+    val skipRun = schedule match {
+      case Some(schedule) => schedule.concurrencyPolicy == ConcurrencyPolicy.Forbid && runsForJob(spec.id).nonEmpty
+      case None           => false
+    }
+    if (skipRun) {
+      log.debug(s"SkippingB Scheduled run for ${spec.id} based on concurrency policy")
```
timcharper: silently dropped is fine?

jobs/src/main/scala/dcos/metronome/jobspec/impl/JobSpecSchedulerActor.scala
```diff
@@ -49,10 +50,26 @@ class JobSpecSchedulerActor(
 
   def runJob(schedule: ScheduleSpec): Unit = {
     log.info(s"Start next run of job ${spec.id}, which was scheduled for $scheduledAt")
-    runService.startJobRun(spec, Some(schedule.startingDeadline))
+    runService.startJobRun(spec, Some(schedule), Some(schedule.startingDeadline))
+    //    if (!(schedule.concurrencyPolicy == ConcurrencyPolicy.Forbid && activeRunsForJobSpec())) {
```
timcharper: should probably get rid of this comment

jobs/src/main/scala/dcos/metronome/jobspec/impl/JobSpecSchedulerActor.scala
```diff
@@ -49,10 +50,26 @@ class JobSpecSchedulerActor(
...
+    //    } else {
+    //      log.debug(s"Skipping Scheduled run for ${spec.id} based on concurrency policy")
+    //    }
     scheduleNextRun()
   }
 
+  def activeRunsForJobSpec(): Boolean = {
```
timcharper: This method is no longer used... remove?

jobs/src/main/scala/dcos/metronome/jobrun/JobRunService.scala
```diff
@@ -41,7 +41,7 @@ trait JobRunService {
     * @param jobSpec the specification to run.
     * @return the started job run
     */
-  def startJobRun(jobSpec: JobSpec, startingDeadline: Option[Duration] = None): Future[StartedJobRun]
+  def startJobRun(jobSpec: JobSpec, schedule: Option[ScheduleSpec] = None, startingDeadline: Option[Duration] = None): Future[StartedJobRun]
```
alenkacz: if we go this way, we should get rid of startingDeadline as a separate field, because then it's also part of schedulespec

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunServiceActor.scala
alenkacz: I would log this as info

Also this is a typo SkippingB
And I would start the word Scheduled with lovercase letter

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunServiceActor.scala
```diff
@@ -42,31 +41,40 @@ class JobRunServiceActor(
...
   def runsForJob(jobId: JobId): Iterable[StartedJobRun] = allJobRuns.values.filter(_.jobRun.id.jobId == jobId)
 
-  def triggerJobRun(spec: JobSpec, startingDeadline: Option[Duration]): Unit = {
+  def triggerJobRun(spec: JobSpec, schedule: Option[ScheduleSpec], startingDeadline: Option[Duration]): Unit = {
     log.info(s"Trigger new JobRun for JobSpec: $spec")
-    val jobRun = JobRun(JobRunId(spec), spec, JobRunStatus.Initial, clock.now(), None, startingDeadline, Map.empty)
-    val startedJobRun = startJobRun(jobRun)
-    sender() ! startedJobRun
+
+    val skipRun = schedule match {
```
alenkacz: What do you think about this:

`val skipRun = schedule.exists(schedule => schedule.concurrencyPolicy == ConcurrencyPolicy.Forbid && runsForJob(spec.id).nonEmpty)`

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunServiceActor.scala
kensipe: I much prefer adding to the logs what this event happened... it is a debug message and not an info

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunServiceActor.scala
kensipe: love it!

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunServiceActor.scala
```diff
@@ -42,31 +41,41 @@ class JobRunServiceActor(
...
-  def triggerJobRun(spec: JobSpec, startingDeadline: Option[Duration]): Unit = {
+  def triggerJobRun(spec: JobSpec, schedule: Option[ScheduleSpec]): Unit = {
     log.info(s"Trigger new JobRun for JobSpec: $spec")
-    val jobRun = JobRun(JobRunId(spec), spec, JobRunStatus.Initial, clock.now(), None, startingDeadline, Map.empty)
-    val startedJobRun = startJobRun(jobRun)
-    sender() ! startedJobRun
+
+    val skipRun = schedule.exists(schedule => schedule.concurrencyPolicy == ConcurrencyPolicy.Forbid && runsForJob(spec.id).nonEmpty)
+    if (skipRun) {
+      log.debug(s"Skipping scheduled run for ${spec.id} based on concurrency policy")
```
alenkacz: I would log this as info

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunServiceActor.scala
```diff
@@ -42,31 +41,41 @@ class JobRunServiceActor(
...
+    } else {
+      val startingDeadline: Option[Duration] = schedule match {
```
alenkacz: schedule.map(_.startingDeadline)

jobs/src/test/scala/dcos/metronome/jobspec/impl/JobSpecSchedulerActorTest.scala
```diff
@@ -67,6 +67,21 @@ class JobSpecSchedulerActorTest extends TestKit(ActorSystem("test")) with FunSui
     system.stop(actor)
   }
 
+  test("If the next scheduled time has reached AND FORBID, a new job run is NOT triggered however a new time is scheduled") {
```
alenkacz: is this test still passing? it"s weird because we moved the logic to other actor...

jobs/src/test/scala/dcos/metronome/jobrun/JobRunServiceFixture.scala
```diff
@@ -32,7 +32,11 @@ object JobRunServiceFixture {
     override def listRuns(filter: (JobRun) => Boolean): Future[Iterable[StartedJobRun]] = {
       Future.successful(specs.values.filter(r => filter(r.jobRun)))
     }
-    override def startJobRun(jobSpec: JobSpec, startingDeadline: Option[Duration] = None): Future[StartedJobRun] = {
+    override def startJobRun(jobSpec: JobSpec, schedule: Option[ScheduleSpec] = None): Future[StartedJobRun] = {
+      val startingDeadline: Option[Duration] = schedule match {
```
alenkacz: schedule.map(_.startingDeadline)

PR #185, opened 2018-01-23
Add authorization to queue endpoint

alenkacz:
Summary:
We should show only jobs user is authorized to see.

JIRA issues: METRONOME-211

### DIFF
--- a/api/src/main/scala/dcos/metronome/api/Authorization.scala
+++ b/api/src/main/scala/dcos/metronome/api/Authorization.scala
@@ -4,7 +4,7 @@ package api
 import akka.util.ByteString
 import dcos.metronome.jobinfo.JobSpecSelector
 import dcos.metronome.jobrun.StartedJobRun
-import dcos.metronome.model.{ JobRun, JobSpec }
+import dcos.metronome.model.{ JobRun, JobSpec, QueuedJobRunInfo }
 import mesosphere.marathon.plugin.auth._
 import mesosphere.marathon.plugin.http.{ HttpRequest, HttpResponse }
 import play.api.http.{ HeaderNames, HttpEntity, Status }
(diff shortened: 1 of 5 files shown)


jobs/src/main/scala/dcos/metronome/queue/impl/LaunchQueueServiceImpl.scala
```diff
@@ -11,6 +11,6 @@ import scala.collection.immutable.Seq
 class LaunchQueueServiceImpl(launchQueue: LaunchQueue) extends LaunchQueueService {
 
   override def list(): Seq[QueuedJobRunInfo] = {
-    launchQueue.list.filter(_.inProgress == true).map(_.toModel)
+    launchQueue.list.filter(_.inProgress).map(_.toModel)
```
kensipe: nice fix.. thanks!

api/src/main/scala/dcos/metronome/api/Authorization.scala
```diff
@@ -38,9 +38,11 @@ case class AuthorizedRequest[Body](identity: Identity, request: Request[Body], a
   def isAllowed(jobSpec: JobSpec): Boolean = authorizer.isAuthorized(identity, ViewRunSpec, jobSpec)
   def isAllowed(jobRun: JobRun): Boolean = isAllowed(jobRun.jobSpec)
   def isAllowed(started: StartedJobRun): Boolean = isAllowed(started.jobRun.jobSpec)
+  // QueuedJobRunInfo extends RunSpec so we can pass it directly to Authorizer
+  def isAllowed(queuedJobRunInfo: QueuedJobRunInfo): Boolean = authorizer.isAuthorized(identity, ViewRunSpec, queuedJobRunInfo)
 
   object Authorized {
-    def unapply(jobSpec: JobSpec): Option[JobSpec] = if (isAllowed(jobSpec)) Some(jobSpec) else None
+    def unapply(jobSpec: JobSpec): Option[JobSpec] = Some(jobSpec).filter(isAllowed)
```
meichstedt: Is this extractor actually used somewhere? Don't see the usage in the diff but maybe it's somewhere else. If it is, never mind. Was just wondering.

alenkacz: I just saw a refactoring suggestion by InteliJ on this line so I refactored it. So it was not directly related to this PR...

PR #186, opened 2018-01-24
Enable SI Tests for current Master

kensipe:
Enable SI Tests for current Master

Summary:
Master updates should have all DCOS environment run an SI tests.
https://jira.mesosphere.com/browse/METRONOME-202

JIRA issues:  METRONOME-202


alenkacz:
@kensipe I would prefer to review with the scripts that are actually updating that PR, because even merging this right now I think does not make much sense...

kensipe:
scripts for working with github... coming soon

PR #189, opened 2018-01-30
Add slack notification when the test pipeline fails

alenkacz:
Summary:
The SI tests in Metronome seems stable enough so I think we can do this. This is just a proposal, feel free to comment.

JIRA issues:


Jenkinsfile.open.si
```diff
@@ -26,7 +26,14 @@ node('py36') {
...
           }
-        } finally {
+        } catch (Exception e) {
+            currentBuild.result = 'FAILURE'
+            slackSend(
+                message: "(;¬_¬) SI tests for Metronome (Open) failed in build `${env.BUILD_NUMBER}`. (<${env.BUILD_URL}|Open>)",
+                color: "danger",
+                channel: "#metronome",
+                tokenCredentialId: "f430eaac-958a-44cb-802a-6a943323a6a8")
+            } finally {
```
kensipe: this doesn't seem properly aligned.

PR #194, opened 2018-02-02
Fix METRONOME-100 regression SI test

alenkacz:
Summary:

- shakedown.masters is actually true if masters count is greater than that number, so not what I wanted
- we need to skip the test until METRONOME-100 fix is in testing/master or until METRONOME-202 is done

JIRA issues: METRONOME-100

### DIFF
--- a/tests/system/test_root_metronome.py
+++ b/tests/system/test_root_metronome.py
@@ -10,10 +10,9 @@ import common
 import shakedown
 import pytest
 
-from common import job_no_schedule, schedule
+from common import job_no_schedule, schedule, not_required_masters_exact_count # NOQA F401
 from dcos import metronome
 from retrying import retry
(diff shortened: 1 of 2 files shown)


tests/system/test_root_metronome.py
```diff
@@ -240,7 +239,8 @@ def test_docker_job():
         assert len(client.get_runs(job_id)) == 1
 
 
-@shakedown.masters(1)
```
zen-dog: we should probably also rename those annotations @kensipe to reflect the logic e.g.  `@shakedown.master_bigger_than` (which is not a great name)

alenkacz: I am afraid that's a backward incompatible change :( if marathon is the only one using shakedown then it's probable possible...

PR #195, opened 2018-02-12
Wait for metronome, not mesos after shutdown

alenkacz:
Summary:
It was supposed to be fix for multimaster cluster, now it's just probably a PR that will make the debugging easier by better error messages. Also now we are waiting for Metronome, not just Mesos (which can make a difference but might now).

JIRA issues: METRONOME-230

### DIFF
--- a/tests/system/common.py
+++ b/tests/system/common.py
@@ -8,6 +8,7 @@ import time
 import pytest
 
 from dcos import metronome
+from retrying import retry
 
 
 def job_no_schedule(id='pikachu', cmd='sleep 10000'):
@@ -88,20 +89,23 @@ def wait_for_mesos_endpoint(timeout_sec=timedelta(minutes=5).total_seconds()):
     return shakedown.time_wait(lambda: shakedown.mesos_available_predicate(), timeout_seconds=timeout_sec)
 
 
-def wait_for_metronome(timeout_sec=timedelta(minutes=5).total_seconds()):
+def ignore_exception(exc):
+    """Used with @retrying.retry to ignore exceptions in a retry loop.
+       ex.  @retrying.retry( retry_on_exception=ignore_exception)
+       It does verify that the object passed is an exception
+    """
+    return isinstance(exc, Exception)
+
+
+@retry(wait_exponential_multiplier=1000, wait_exponential_max=5*60*1000, retry_on_exception=ignore_exception) # 5 minutes
+def wait_for_metronome():
     """ Waits for the Metronome API url to be available for a given timeout. """
-
-    shakedown.time_wait(lambda: metronome_available_predicate(),
-                        timeout_seconds=timeout_sec)
-
-
-def metronome_available_predicate():
     url = metronome_api_url()
     try:
         response = http.get(url)
-        return response.status_code == 200
+        assert response.status_code == 200, f"Expecting Metronome service to be up but it did not get healthy after 5 minutes. Last response: {response}"
     except Exception as e:
-        return False
+        assert False, f"Expecting Metronome service to be up but it did not get healthy after 5 minutes. Last exception: {e}"
 
 
 def metronome_api_url():
--- a/tests/system/common.py
+++ b/tests/system/common.py
@@ -88,6 +88,19 @@ def wait_for_mesos_endpoint(timeout_sec=timedelta(minutes=5).total_seconds()):
     return shakedown.time_wait(lambda: shakedown.mesos_available_predicate(), timeout_seconds=timeout_sec)
 
 
+def wait_for_metronome(timeout_sec=timedelta(minutes=5).total_seconds()):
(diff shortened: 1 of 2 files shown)


tests/system/common.py
```diff
@@ -88,6 +88,19 @@ def wait_for_mesos_endpoint(timeout_sec=timedelta(minutes=5).total_seconds()):
...
 
+def wait_for_metronome(timeout_sec=timedelta(minutes=5).total_seconds()):
+    """ Waits for the Metronome API url to be available for a given timeout.
+        It needs to succeed as many times as we have masters because that means that all adminrouters on all masters
+        are probably aware of the restarted leader.
+        Without this in multimaster setup this can be very flaky.
+    """
+
+    master_count = len(shakedown.get_all_masters())
+    shakedown.time_wait(lambda: metronome_available_predicate(),
```
jeschkies: I'd much rather use retry with assert since it is easier to debug an `AssertException: assert 2 == 3` than a `shakedown.TimeoutException`. See [MARATHON-8077](https://jira.mesosphere.com/browse/MARATHON-8077). I'm not too strong on asserts but find it the easiest way.

kensipe: we use shakedown.. and there hasn't been a stated direction to move away from it.   If there is an issue... shouldn't we fix the issue with shakedown?

jeschkies: Well, that's another debate from what I perceive. There is also dcos-test-utils which ported some things from shakedown and is not dependent on the dcos-cli as Shakedown is.

In any case, you raise a valid point. Where should we add improvements to our error reporting?

tests/system/common.py
```diff
@@ -88,6 +88,19 @@ def wait_for_mesos_endpoint(timeout_sec=timedelta(minutes=5).total_seconds()):
...
+                        timeout_seconds=timeout_sec,
+                        required_consecutive_success_count=master_count)
```
jeschkies: Hm, why do we retry as many times as we have masters?

kensipe: this is important given a frontend ELB (which is the case with AWS), which will cycle through the masters...    There are cases where you need to make sure all the masters are recovered to be in an expected working condition...   the biggest issue you are over coming by making sure that all the masters are responding ok... is to avoid flaky tests... you don't have control over which master the next call will be routed to... so if 2 out of 3 are good... call 1 works, call 2 works and call 3 fails but only because the master isn't fully up.   3 "consecutive" successes in a round robin LB model will make sure that all 3 masters are up.

jeschkies: Why not check each master then? It seems like some black magic to rely on ELB cycling through all masters.

tests/system/common.py
jeschkies: I see your point. However, I'd like to avoid going down a road we may not want to be on. What's the benefit aside from consistency? I think there is a better alternative right now without much effort.

tests/system/test_root_metronome.py
```diff
@@ -280,4 +279,7 @@ def job(job_json):
     try:
         yield
     finally:
-        client.remove_job(job_id, True)
+        try:
```
alenkacz: this took me ~1h to figure out :D because the test was still not giving me the nice assert failures and it was because when this thing failed, it printed the dcos error and not the nice assertion error...

alenkacz: btw. we do the same thing with exceptions in Marathon cleanup methods

tests/system/common.py
```diff
@@ -88,13 +89,23 @@ def wait_for_mesos_endpoint(timeout_sec=timedelta(minutes=5).total_seconds()):
...
+
+
+@retry(wait_exponential_multiplier=1000, wait_exponential_max=5*60*1000, retry_on_exception=ignore_exception) # 5 minutes
+def wait_for_metronome():
+    """ Waits for the Metronome API url to be available for a given timeout. """
     url = metronome_api_url()
     try:
         response = http.get(url)
-        return response.status_code == 200
+        assert response.status_code == 200, f"Expecting Metronome service to be up but it did not get healthy after 5 minutes. Last response: {response}"
```
jeschkies: If this is a Requests response it probably should be `response.content` or `response.text`.

tests/system/common.py
```diff
@@ -88,13 +89,23 @@ def wait_for_mesos_endpoint(timeout_sec=timedelta(minutes=5).total_seconds()):
...
     except Exception as e:
-        return False
+        assert False, f"Expecting Metronome service to be up but it did not get healthy after 5 minutes. Last exception: {e}"
```
jeschkies: Hm, this seems a little odd. We could throw an exception with the original exception as its cause. This would take it too far though.

tests/system/test_root_metronome.py
jeschkies: You mean it printed the error of the cleanup? Good catch!

PR #196, opened 2018-02-13
Clarify the /runs endpoint

alenkacz:
JIRA issues: METRONOME-182

### DIFF
--- a/api/src/main/resources/public/api/v1/jobs.raml
+++ b/api/src/main/resources/public/api/v1/jobs.raml
@@ -276,7 +276,7 @@ post:
     #
     get:
       description:
-        Get the list of all runs for this jobId.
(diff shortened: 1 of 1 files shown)


alenkacz:
low hanging 🍎

api/src/main/resources/public/api/v1/jobs.raml
```diff
@@ -276,7 +276,7 @@ post:
     #
     get:
       description:
-        Get the list of all runs for this jobId.
+        Get the list currently active runs for this jobId. Once the run is finished, it will not be available through this endpoint.
```
meln1k: `Get the list currently active` =>  `Get the list of currently active`

meichstedt: agreed

alenkacz: thanks!

PR #197, opened 2018-02-14
HTTP 500 Errors Based on ZK Storage

kensipe:
HTTP 500 Errors Based on ZK Storage

Summary:
The `LoadContentOnStartup` trait has a `loadAll` function in it's `preStart`.   It is possible for the  `loadAll` function to throw a `StoreCommandFailedException` (which is in the documentation for the repo.get).   This caused the `JobRunServiceActor` which is of type LoadContentOnStart to fail to start.   This caused controller calls for JobSpecs (such as get a list of jobspecs) to timeout since the actor never responded to a request for 10s.  In this state, Metronome would never work correctly until the zk was cleared.

There are 2 reasonable reasons for this situation to occur:

1. The zk path for `job-spec`/some-job-id was created, however the process was killed prior to the job spec data being added to the zk path.  We do NOT use zk transactions.  The twitter zk lib we use does not expose them.
2. The data stored at a zk path is corrupt.

The change included in this PR provides error handling around data loading.   It will "skip" dangling paths or paths with corrupt data.

https://jira.mesosphere.com/browse/METRONOME-218

JIRA issues:   METRONOME-218

### DIFF
--- a/jobs/src/main/scala/dcos/metronome/repository/LoadContentOnStartup.scala
+++ b/jobs/src/main/scala/dcos/metronome/repository/LoadContentOnStartup.scala
@@ -2,6 +2,7 @@ package dcos.metronome
 package repository
 
 import akka.actor.{ Actor, ActorLogging, Stash }
+import mesosphere.marathon.StoreCommandFailedException
 
 import scala.concurrent.Future
 import scala.util.{ Failure, Success }
@@ -33,7 +34,7 @@ trait LoadContentOnStartup[Id, Model] extends Actor with Stash with ActorLogging
     val loadAllFuture = repo.ids().flatMap { ids =>
       ids.foldLeft(Future.successful(List.empty[Model])) {
         case (resultsFuture, id) => resultsFuture.flatMap { res =>
-          repo.get(id).map(_.map(_ :: res).getOrElse(res))
+          getModel(id).map(_.map(_ :: res).getOrElse(res))
         }
       }
     }
@@ -43,6 +44,14 @@ trait LoadContentOnStartup[Id, Model] extends Actor with Stash with ActorLogging
       case Failure(ex) =>
         log.error(ex, "Can not load initial data. Give up.")
         throw ex
+    }
+  }
+
+  private def getModel(id: Id): Future[Option[Model]] = {
+    repo.get(id).recoverWith {
+      case ex: StoreCommandFailedException =>
+        log.error(s"ID $id is a dangling path with data recovery issues.  Exception message: ${ex.getMessage}")
+        Future { None }
     }
   }
 }


ichernetsky:
I think we removed this masking/ignoring of store failures in Marathon, and I do not see the point of introducing it here. It might be irritating to have dangling nodes, yes, but returning `None` in case of a real ZK failure could lead to data loss/overwriting. @kensipe @timcharper what do you think?

jobs/src/main/scala/dcos/metronome/repository/LoadContentOnStartup.scala
```diff
@@ -45,6 +46,14 @@ trait LoadContentOnStartup[Id, Model] extends Actor with Stash with ActorLogging
         throw ex
     }
   }
+
+  private def getModel(id: Id): Future[Option[Model]] = {
+    repo.get(id).recoverWith {
+      case ex: StoreCommandFailedException =>
+        log.error(s"ID $id is a dangling path with data recovery issues.  Exception message: ${ex.getMessage}")
+        Future { None }
```
meln1k: What do you think about `Future.successful(None)`?

kensipe:
@ichernetsky yeah... I had some concerns... That is why I was specific around `StoreCommandFailedException` only...   Other exception failures would still result in the previous condition.

kensipe:
For any other failure... we should abort the process IMO.

ichernetsky:
I checked the Metronome code and the Marathon code, Metronome depends on. Metronome uses ZK store implementation from Marathon. This store masks some exceptions (this masking got removed in 1.4.x+), and the store converts all `KeeperException`s into `StoreCommandFailedException`s. So ignoring `StoreCommandFailedException` will mask all kinds of `KeepException`s.

alenkacz:
hmm @ichernetsky from what you wrote the proper solution then might be to actually update to the newest Marathon and then maybe do some changes ... ? Do I understand it correctly?

ichernetsky:
@alenkacz, Ken and I discussed it privately in Slack, and we concluded that this might be OK in case of Metronome because its data layout in ZK is much simpler if comparing it with Marathon's one. So, before merging this PR someone needs to build a detailed mental model of Metronome data layout and all operations on it. I would like to underline that it _might_ be OK. On the other hand, even it turns out that it is more-or-less OK, the way it is currently done in this PR will mask all kinds of exceptions, not just the ones like `KeeperException.NoNodeException`. So in order to address the issue of dangling nodes, instead of returning `None` on all `StoreCommandFailedException`, it probably makes sense to check its cause `e.getCause()`, and if the cause is `NoNodeException`, then it returns just `None`. What do you think?

jobs/src/main/scala/dcos/metronome/repository/LoadContentOnStartup.scala
```diff
@@ -45,6 +48,24 @@ trait LoadContentOnStartup[Id, Model] extends Actor with Stash with ActorLogging
...
     }
   }
+
+  private def getModel(id: Id): Future[Option[Model]] = {
+    repo.get(id).recoverWith {
+      case ex: StoreCommandFailedException =>
+        ex.getCause match {
+          case cause: InvalidProtocolBufferException =>
+            log.error(s"ID $id has corrupt data and will be ignored.  Zk will need to be manually repaired.  Exception message: ${cause.getMessage}")
+            Future.successful(None)
```
timcharper: I think that we should throw an exception in this case, don't you?

kensipe: @timcharper  No!   We can't!   This is in the PreStart of the Actor... the exception is what is taking out the Actor and causing the 500 errors.      this approach shouldn't be an issue... there will be no data loss and they can't write over the existing node.

timcharper: I'm really confused. Why wouldn't we crash in this case, if zookeeper data is detected to be corrupt?

timcharper: > Seems appropriate for `log.error(ex, "Can not load initial data. Give up.")` to end up with a process abortion.   The invalid state of zk should have completely stopped the process.

kensipe: that is an older message.. and may apply sometimes... you and I discussed this and I thought we landed here...

kensipe: my bad...   we do want to die for corrupt data...   we want to prune for no node expection... I'll have an update shortly

PR #198, opened 2018-02-15
Do not throw exception for known error

alenkacz:
Summary:
Low hanging 🍎 

JIRA issues: METRONOME-197

### DIFF
--- a/api/src/main/scala/dcos/metronome/api/v1/controllers/JobScheduleController.scala
+++ b/api/src/main/scala/dcos/metronome/api/v1/controllers/JobScheduleController.scala
@@ -76,12 +76,15 @@ class JobScheduleController(
     }
 
     withJobSpec(id) { spec =>
-      jobSpecService.updateJobSpec(id, deleteSchedule)
-        .map(_ => Ok)
-        .recover{
-          case JobSpecDoesNotExist(_)       => NotFound(UnknownJob(id))
-          case ex: IllegalArgumentException => NotFound(UnknownSchedule(scheduleId))
-        }
+      if (spec.schedules.count(_.id == scheduleId) != 1) {
+        Future.successful(NotFound(UnknownSchedule(scheduleId)))
+      } else {
+        jobSpecService.updateJobSpec(id, deleteSchedule)
+          .map(_ => Ok)
+          .recover {
+            case JobSpecDoesNotExist(_) => NotFound(UnknownJob(id))
+          }
+      }
     }
   }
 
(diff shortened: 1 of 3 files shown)


api/src/main/scala/dcos/metronome/api/v1/controllers/JobScheduleController.scala
```diff
@@ -76,12 +76,15 @@ class JobScheduleController(
     }
 
     withJobSpec(id) { spec =>
```
alenkacz: the original IllegalArgumentException was throws few lines above here via a require method, just unwrap more of the code when reviewing :)

api/src/main/scala/dcos/metronome/api/v1/controllers/JobScheduleController.scala
```diff
@@ -76,12 +76,15 @@ class JobScheduleController(
...
-      jobSpecService.updateJobSpec(id, deleteSchedule)
-        .map(_ => Ok)
-        .recover{
-          case JobSpecDoesNotExist(_)       => NotFound(UnknownJob(id))
```
meichstedt: The impl of this controller is pretty weird: `withJobSpec` reads the jobSpec, and will result in a `JobSpecDoesNotExist` exception if it doesn't exist. In case we could fetch the jobSpec, we'll call `updateJobSpec`, which will basically do the same – try to fetch the spec, apply a function or throw `JobSpecDoesNotExist`.

The impl could simply _try_ to create/update/delete the spec and handle the exceptions accordingly. There's quite some DRY's in here that _could_ be resolved.

api/src/main/scala/dcos/metronome/api/v1/controllers/JobScheduleController.scala
meichstedt: I don't have a nice solution at hand but I think the flow could be simpler. We already pass along a function that performs the create/update/delete, we could as well wrap the predicate (must not exist yet/must exist etc) and pass that along.

Ideally we'd only do `jobSpecService.updateJobSpec` and handle either _JobSpec doesn't exist_ or _operation on schedule is not applicable_.

alenkacz:
@meichstedt I did some refactoring...

What I realized is that it's actually hard to get rid of the withJobSpec because what it is actually doing is authorization and we probably don't want to move that to the service layer (do we?)

But I got rid of the require and renamed the method...

If querying jobspec more than once is actually a bottle neck (or might potentially be?) we might also introduce a service method that accepts jobspec as parameter - right now what remains is that we query that thing multiple times

PR #202, opened 2018-02-27
Additional cron validation

alenkacz:
Summary:
Submitting invalid CRON like '0 0 31 2 *' (run on 31.2.) basically made the whole Metronome unresponsive. Core of this issue lies in the cron-utils library we use.
It does some basic validation of cron fields but not anything on this level.
As a result, when we call nextExecution, there is a while loop looking for a next schedule that spins up the CPU and keeps the whole system busy (maybe even never terminating).
The proper fix for this is patching cron-utils (adding validation as well as fixing the nextExecution method).
I would do that ^^^ but I first implemented the validation logic inside Metronome itself (it uses the same abstraction as will be used inside the library) and I also added tests proving that it works now :-) (e.g. the controller test was timing out prior to my fix)

JIRA issues: METRONOME-236

### DIFF
--- a/api/src/test/scala/dcos/metronome/api/v1/CronSpecFormatTest.scala
+++ b/api/src/test/scala/dcos/metronome/api/v1/CronSpecFormatTest.scala
@@ -2,7 +2,7 @@ package dcos.metronome
 package api.v1
 
 import dcos.metronome.api.v1.models._
-import dcos.metronome.model.CronSpec
+import dcos.metronome.model.{ CronSpec, CronSpecValidation }
 import dcos.metronome.utils.test.Mockito
 import org.joda.time.DateTime
 import org.scalatest.{ FunSuite, Matchers }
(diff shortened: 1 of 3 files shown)


api/src/test/scala/dcos/metronome/api/v1/CronSpecFormatTest.scala
```diff
@@ -156,4 +156,36 @@ class CronSpecFormatTest extends FunSuite with Mockito with Matchers {
...
   }
+
+  test("Invalid cron because february cannot have 31 days") {
+    val cronString = "0 0 31 2 *"
+
+    val thrown = the[IllegalArgumentException] thrownBy CronSpec(cronString)
+    thrown.getMessage should include(CronSpecValidation.validDayOfMonth)
+  }
+
+  test("Invalid cron because february will never be in a range 30-31") {
```
kensipe: it would be great to have a feb 29 date test

jobs/src/main/scala/dcos/metronome/model/CronSpec.scala
```diff
@@ -70,3 +77,44 @@ object CronSpec {
     }
   }
 }
+
+object CronSpecValidation {
+  val validDayOfMonth = "Day of the month must exist in the provided month and year (e.g. February have only < 29 days so running cron on 30.2. is invalid)"
```
kensipe: several nits here:
1. the message is incorrect... it doesn't have anything to do with the year... This is valid day of the month or within the possible days of a given month.
2. Feb has < 30... or =< 29.   29 is a valid day of feb in leap years
3. I don't like the notation of 30.2.   we don't need to "americanize" it... but it is made more clear to write Feb 30 IMO.   There isn't room for interpretation.

jobs/src/main/scala/dcos/metronome/model/CronSpec.scala
```diff
@@ -70,3 +77,44 @@ object CronSpec {
...
+}
+
+/**
+  * In the existing validation logic inside cron-utils there is not validation that the given day exist in a month.
```
kensipe: this sentence is hard to read IMO.    Perhaps . "Cron-utils existing validation logic does NOT validate that a given day is possible for a given month."

jobs/src/main/scala/dcos/metronome/model/CronSpec.scala
```diff
@@ -70,3 +77,44 @@ object CronSpec {
...
+  * This validator covers that use case disallowing schedules like 0 0 31 2 * (run on 31.2.)
```
kensipe: nit:  `31.2` as stated above.

jobs/src/main/scala/dcos/metronome/model/CronSpec.scala
```diff
@@ -70,3 +77,44 @@ object CronSpec {
...
+  */
+class CronDaysInMonthValidation extends CronConstraint(CronSpecValidation.validDayOfMonth) {
+  def daysExistInAMonths(days: Seq[Int], months: Seq[Int]): Boolean = {
+    months.exists(m => days.exists(d => dayExistInAMonth(d, m)))
+  }
+  val maxDaysOfMonth = Array(31, 29, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31)
+  def dayExistInAMonth(day: Int, month: Int): Boolean = {
+    day < maxDaysOfMonth(month - 1)
```
kensipe: `day <= maxDaysOfMonth(month - 1)` ?

kensipe:
this should pass... 

```
  test("Not fail when leap year is possible") {
    val cronString = "0 0 29 2 *"

    val spec = CronSpec(cronString)
    val currentDateTime: DateTime = new DateTime(2017, 10, 2, 10, 0)

    val nextCronDate = spec.nextExecution(currentDateTime)

    nextCronDate shouldEqual new DateTime(2020, 2, 29, 0, 0)
  }

```

and does if `CronSpecValidation` is changed to: `day <= maxDaysOfMonth(month - 1)`

jobs/src/main/scala/dcos/metronome/model/CronSpec.scala
alenkacz: oh god, such a stupid mistake. Thanks :)!!! I refactored this thing so many times that it just got removed at one time :D

alenkacz:
thanks for the fixes @kensipe

PR #204, opened 2018-02-28
Fix flaky tests on 0.3 branch

kensipe:
tests should not fail based on actor timeouts

Summary:
tests should not fail based on actor timeouts

JIRA issues:


alenkacz:
@kensipe looking at the failing test in this PR, I think we need to backport my test fixes to this branch... I"ll try to do that during this day.

Thanks for the patience config fix!

PR #211, opened 2018-03-16
Bump version of play to 2.6

alenkacz:
Summary:
This is needed to be able to go to scala 2.12.

JIRA issues: METRONOME-244

PR #212, opened 2018-03-17
Update Metronome to Marathon 1.6.322

alenkacz:
This branch merges into `karsten/bump-play` (so it is built on top of play 2.6 update). I wanted these two to be reviewed separately...

Depends on #211.

JIRA issues: METRONOME-244

### DIFF
--- a/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
@@ -129,7 +129,7 @@ class JobRunExecutorActor(
   def tasksFromInstanceTracker(): Iterable[JobRunTask] = {
     instanceTracker.specInstancesSync(runSpecId).map(a => a.appTask).collect {
       case task: Task => JobRunTask(task)
-      case task => throw UnexpectedTaskState(task)
+      case task       => throw UnexpectedTaskState(task)
     }
   }
 
--- a/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
@@ -55,7 +55,7 @@ class JobRunExecutorActor(
       case JobRunStatus.Initial =>
         becomeCreating()
       case JobRunStatus.Starting | JobRunStatus.Active =>
-        val existingTasks = tasksFromTaskTracker()
+        val existingTasks = tasksFromInstanceTracker()
         existingTasks match {
           case tasks if tasks.nonEmpty =>
             // the actor was probably just restarted and we did not lose contents of the launch queue
@@ -126,7 +126,7 @@ class JobRunExecutorActor(
     startingDeadlineTimer = None
   }
 
-  def tasksFromTaskTracker(): Iterable[JobRunTask] = {
+  def tasksFromInstanceTracker(): Iterable[JobRunTask] = {
     instanceTracker.specInstancesSync(runSpecId).map(a => a.appTask).collect {
       case task: Task => JobRunTask(task)
       case task => throw UnexpectedTaskState(task)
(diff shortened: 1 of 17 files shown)


jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -129,8 +128,8 @@ class JobRunExecutorActor(
 
   def tasksFromTaskTracker(): Iterable[JobRunTask] = {
     instanceTracker.specInstancesSync(runSpecId).map(a => a.appTask).collect {
-      case task: LaunchedEphemeral => JobRunTask(task)
-      case task: Task              => throw UnexpectedTaskState(task)
+      case task: Task => JobRunTask(task)
+      case task: Task => throw UnexpectedTaskState(task)
```
jeschkies: I don't think this works. Did you mean `case _ => throw UnexpectedTaskState()`?

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -129,8 +128,8 @@ class JobRunExecutorActor(
 
   def tasksFromTaskTracker(): Iterable[JobRunTask] = {
     instanceTracker.specInstancesSync(runSpecId).map(a => a.appTask).collect {
-      case task: LaunchedEphemeral => JobRunTask(task)
-      case task: Task              => throw UnexpectedTaskState(task)
+      case task: Task => JobRunTask(task)
```
meichstedt: Nit: rename the def `tasksFromInstanceTracker`?

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -129,8 +128,8 @@ class JobRunExecutorActor(
...
+      case task => throw UnexpectedTaskState(task)
```
meln1k: It seems like a breaking change. Are we sure it should be done this way>

alenkacz: LaunchedEphemeral was removed in 1.6 ... I am not sure if this is breaking or not. @meichstedt

meichstedt: No that's fine. Before 1.6, we had a `Task` trait and subclasses (`LaunchedEphemeral` and others) to that. 1.6 removed the trait so everything is a Task case class now. The only subclass that would've been a problem was `Reserved` since that's not a launched task but only a reservation representation.

kensipe:
looks related to:  https://jira.mesosphere.com/browse/MARATHON-7971
It is solved in marathon with https://github.com/dcos/dcos/blob/master/packages/marathon/build#L45
however play has a different startup

kensipe:
```
[error] p.c.s.AkkaHttpServer - Cannot load SSL context
java.lang.reflect.InvocationTargetException: null
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at play.core.server.ssl.ServerSSLEngine$.createScalaSSLEngineProvider(ServerSSLEngine.scala:103)
	at play.core.server.ssl.ServerSSLEngine$.createSSLEngineProvider(ServerSSLEngine.scala:35)
	at play.core.server.AkkaHttpServer.$anonfun$httpsServerBinding$1(AkkaHttpServer.scala:126)
	at play.core.server.AkkaHttpServer.$anonfun$httpsServerBinding$1$adapted(AkkaHttpServer.scala:124)
	at scala.Option.map(Option.scala:146)
	at play.core.server.AkkaHttpServer.<init>(AkkaHttpServer.scala:124)
Caused by: java.nio.file.AccessDeniedException: /opt/mesosphere/packages/metronome--962c868a74afcf2f0bbb4cb76b8958adcb1982e7/conf/generated.keystore
```

jeschkies:
@kensipe what should we do with #211? Can we merge that one?

jeschkies:
WE could also add a smoke test for start up.

kensipe:
@jeschkies I'll run tests on #211 and merge on pass 
(good point!)

PR #213, opened 2018-03-19
Make a script for running metronome locally

alenkacz:
I had to do these steps like 300 times in the last 2 weeks so I finally automated them through this script :)

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -2,7 +2,7 @@
 
 ZK_URL="${1:-zk://127.0.0.1:2181/metronome}"
 MESOS_MASTER_URL="${2:-127.0.0.1:5050}"
-PLAY_PORT="${3:-8888}"
+HTTP_PORT="${3:-9000}"
 
 sbt universal:packageBin
 ZIP_COUNT=`ls target/universal/metronome*.zip | wc -l`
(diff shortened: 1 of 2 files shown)


run.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+ZK_URL="${1:-zk://127.0.0.1:2181/metronome}"
+MESOS_MASTER_URL="${2:-127.0.0.1:5050}"
+PLAY_PORT="${3:-8888}"
```
zen-dog: why don't we call it `HTTP_PORT`? ;) And it `8888` or `8989` a better default one?

alenkacz: good point, thanks

tobilg:
Slightly off-topic, there is also a repo for a Docker build for Metronome at https://github.com/mesoshq/metronome which could also be used locally.

run.sh
```diff
@@ -0,0 +1,16 @@
...
+sbt universal:packageBin
+ZIP_COUNT=`ls target/universal/metronome*.zip | wc -l`
+if [ ${ZIP_COUNT} -gt 1 ]; then
+    echo "Multiple metronome zip files inside /target/universal. Run 'sbt clean' or remove one manually.".
+    exit 1
+fi
+# unpack the package
+unzip -o -d target/universal -a target/universal/metronome-*.zip
+chmod +x target/universal/metronome/metronome-*/bin/metronome
+LIBPROCESS_IP=127.0.0.1 ./target/universal/metronome/metronome-*/bin/metronome -d -v -Dmetronome.framework.name=metronome-dev -Dmetronome.zk.url=$ZK_URL -Dmetronome.mesos.master.url=$MESOS_MASTER_URL -Dplay.server.http.port=$PLAY_PORT
```
kensipe: `./target/universal/metronome/metronome-*/bin/metronome` should be `./target/universal/metronome-*/bin/metronome`

run.sh
kensipe: agree with zen-dog and think the default should be 9000 like it is every where else

kensipe:
also think it should go in a `run` or `script` dir... but isn't that big of a deal.

PR #215, opened 2018-03-29
Support secrets in Metronome

meln1k:
Description: propagation of secrets & serialization tests

JIRA issues: METRONOME-246

### DIFF
--- a/jobs/src/main/scala/dcos/metronome/model/JobSpec.scala
+++ b/jobs/src/main/scala/dcos/metronome/model/JobSpec.scala
@@ -4,8 +4,12 @@ package model
 import com.wix.accord.Validator
 import com.wix.accord.dsl._
 import dcos.metronome.model.JobRunSpec._
+import dcos.metronome.utils.glue.MarathonConversions
 import mesosphere.marathon.api.v2.Validation._
-import mesosphere.marathon.plugin.{ ApplicationSpec, EnvVarValue, NetworkSpec, Secret, VolumeMountSpec, VolumeSpec }
+import mesosphere.marathon
+import mesosphere.marathon.plugin.{ ApplicationSpec, NetworkSpec, Secret, VolumeSpec, VolumeMountSpec }
+
+import scala.collection.immutable._
 
 case class JobSpec(
   id:          JobId,
@@ -17,8 +21,8 @@ case class JobSpec(
 
   override val user: Option[String] = run.user
   override val acceptedResourceRoles: Set[String] = Set.empty
-  override val secrets: Map[String, Secret] = Map.empty
-  override val env: Map[String, EnvVarValue] = mesosphere.marathon.state.EnvVarValue(run.env)
+  override val secrets: Map[String, Secret] = MarathonConversions.secretsToMarathon(run.secrets)
+  override val env: Map[String, marathon.state.EnvVarValue] = MarathonConversions.envVarToMarathon(run.env)
   override val volumes: Seq[VolumeSpec] = Seq.empty
   override val networks: Seq[NetworkSpec] = Seq.empty
   override val volumeMounts: Seq[VolumeMountSpec] = Seq.empty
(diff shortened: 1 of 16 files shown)


jobs/src/main/scala/dcos/metronome/model/JobSpec.scala
```diff
@@ -4,8 +4,12 @@ package model
 import com.wix.accord.Validator
 import com.wix.accord.dsl._
 import dcos.metronome.model.JobRunSpec._
+import dcos.metronome.utils.glue.MarathonConversions
 import mesosphere.marathon.api.v2.Validation._
-import mesosphere.marathon.plugin.{ ApplicationSpec, EnvVarValue, NetworkSpec, Secret, VolumeMountSpec, VolumeSpec }
+import mesosphere.marathon
+import mesosphere.marathon.plugin.{ ApplicationSpec, NetworkSpec, Secret, VolumeSpec, VolumeMountSpec }
+
+import scala.collection.immutable._
```
kensipe: is there a reason this is necessary?   the double package metronome object should use immutable for Seq and IndexedSeq

jobs/src/main/scala/dcos/metronome/model/SecretDef.scala
```diff
@@ -0,0 +1,28 @@
...
+      })
+      if (_errors.nonEmpty) _errors.reduceOption[play.api.libs.json.JsError](_.++(_)).getOrElse(_errors.head)
+      else play.api.libs.json.JsSuccess(SecretDef(source = source.get))
+    }
+    def writes(o: SecretDef): play.api.libs.json.JsValue = {
+      val source = play.api.libs.json.Json.toJson(o.source)
+      play.api.libs.json.JsObject(Seq(("source", source)).filter(_._2 != play.api.libs.json.JsNull).++(Seq.empty))
+    }
+  }
+  val ConstraintSourceMinlength = 1
```
kensipe: upper case L in length.. that is a separate word.   `ConstraintSourceMinLength`

jobs/src/main/scala/dcos/metronome/model/SecretDef.scala
```diff
@@ -0,0 +1,28 @@
...
+  *   The format dependes on the secret store
+  *   minLength: 1
+  */
+case class SecretDef(source: String)
+
+object SecretDef {
+  import play.api.libs.json.Reads._
+  implicit object playJsonFormat extends play.api.libs.json.Format[SecretDef] {
+    def reads(json: play.api.libs.json.JsValue): play.api.libs.json.JsResult[SecretDef] = {
+      val source = json.\("source").validate[String](play.api.libs.json.JsPath.read[String](minLength[String](1)))
```
kensipe: shouldn't the 1 be `ConstraintSourceMinlength` ?   no magic numbers

src/main/resources/application.conf
```diff
@@ -369,7 +369,7 @@ metronome {
   # "vips" (Currently n/a) can be used to enable the networking VIP integration UI.
   # "task_killing" can be used to enable the TASK_KILLING state in Mesos (0.28 or later)
   # "external_volumes" (Currently n/a) can be used if the cluster is configured to use external volumes.
-  # Example: --enable_features vips,task_killing,external_volumes
+  # Example: --enable_features vips,task_killing,external_volumes,secrets
```
kensipe: how is this enabled?   there isn't anything in our code that makes this true

jobs/src/main/scala/dcos/metronome/model/SecretDef.scala
```diff
@@ -0,0 +1,28 @@
+package dcos.metronome.model
+
+/**
+  * A secret declaration
+  * @param source The source of the secrets value.
```
kensipe: this could use improvement... it really just repeats what is already known... the SecretDef is a secret declaration... and the param source is the source of a secret value.        There should be a lot more context here.   It should be clear what the source is.. a reference to a secret which will be injected with a value from the secret store.

meln1k:
pushed to a proper branch: https://github.com/dcos/metronome/pull/216

PR #218, opened 2018-04-06
Job Secrets System Integration 

kensipe:
Job Secrets System Integration 

Summary:
Adding SI test for job secrets for env vars.

JIRA issues:  DCOS_OSS-2338

### DIFF
--- a/tests/system/test_root_metronome.py
+++ b/tests/system/test_root_metronome.py
@@ -257,7 +257,7 @@ def test_secret_env_var(secret_fixture):
             assert len(client.get_runs(job_id)) == 1
             run_id = client.get_runs(job_id)[0]['id']
             stdout, stderr, return_code = shakedown.run_dcos_command("task log {} secret-env".format(run_id))
-            assert  secret_value == stdout.rstrip()
+            assert secret_value == stdout.rstrip()
 
 
 @common.masters_exact(1)
--- a/tests/system/test_root_metronome.py
+++ b/tests/system/test_root_metronome.py
@@ -253,12 +253,13 @@ def test_secret_env_var(secret_fixture):
         client.run_job(job_id)
 
         @retry(wait_fixed=1000, stop_max_delay=5000)
-        def task_exist():
-            assert len(client.get_runs(job_id)) == 1
+        def job_run_has_secret():
+            assert len(client.get_runs(job_id)) == 1, "trigger job should be running"
             run_id = client.get_runs(job_id)[0]['id']
             stdout, stderr, return_code = shakedown.run_dcos_command("task log {} secret-env".format(run_id))
-            assert secret_value == stdout.rstrip()
+            assert secret_value == stdout.rstrip(), "secret value in stdout log Incorrect or missing"
 
+        job_run_has_secret()
 
 @common.masters_exact(1)
 @pytest.mark.skip(reason="we need to wait until METRONOME-100 gets to testing/master")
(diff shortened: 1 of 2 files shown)


tests/system/test_root_metronome.py
```diff
@@ -239,6 +240,26 @@ def test_docker_job():
...
+    secret_name, secret_value = secret_fixture
+
+    client = metronome.create_client()
+    job_id = uuid.uuid4().hex
+    job_def = common.job_with_secrets(job_id, secret_name=secret_name)
+    with job(job_def):
+        client.run_job(job_id)
+
+        @retry(wait_fixed=1000, stop_max_delay=5000)
+        def task_exist():
```
alenkacz: this is just a method definition, right? Don't you have to call this method?

tests/system/test_root_metronome.py
```diff
@@ -239,6 +240,26 @@ def test_docker_job():
...
+            assert len(client.get_runs(job_id)) == 1
+            run_id = client.get_runs(job_id)[0]['id']
+            stdout, stderr, return_code = shakedown.run_dcos_command("task log {} secret-env".format(run_id))
+            assert secret_value == stdout.rstrip()
```
alenkacz: please add some meaningful string description for the asserts to ease the debugging when assert fails

tests/system/test_root_metronome.py
```diff
@@ -239,6 +240,26 @@ def test_docker_job():
...
+
+    client = metronome.create_client()
+    job_id = uuid.uuid4().hex
+    job_def = common.job_with_secrets(job_id, secret_name=secret_name)
+    with job(job_def):
+        client.run_job(job_id)
+
+        @retry(wait_fixed=1000, stop_max_delay=5000)
+        def task_exist():
+            assert len(client.get_runs(job_id)) == 1
```
alenkacz: please add some meaningful string description for the asserts to ease the debugging when assert fails

tests/system/test_root_metronome.py
kensipe: the fact that the secret value doesn't match the secret in the text file isn't meaningful enough?

tests/system/test_root_metronome.py
kensipe: yeah... I'm not sure what happened... on it... THX!!!

tests/system/test_root_metronome.py
alenkacz: for me the debugging would be easier if something like this was there: "Looking for secret value in the task logs. Expecting 'x' but found 'y'"

tests/system/test_root_metronome.py
```diff
@@ -239,6 +240,27 @@ def test_docker_job():
...
+    job_def = common.job_with_secrets(job_id, secret_name=secret_name)
+    with job(job_def):
+        client.run_job(job_id)
+
+        @retry(wait_fixed=1000, stop_max_delay=5000)
+        def job_run_has_secret():
+            assert len(client.get_runs(job_id)) == 1, "trigger job should be running"
+            run_id = client.get_runs(job_id)[0]['id']
+            stdout, stderr, return_code = shakedown.run_dcos_command("task log {} secret-env".format(run_id))
+            assert secret_value == stdout.rstrip(), "secret value in stdout log Incorrect or missing"
```
alenkacz: please also add the value that we found there (so stdout.rstrip())

and Incorrect should be lowercase

kensipe:
failure is unrelated

PR #219, opened 2018-04-09
Info endpoint shows incorrect version of Metronome

kensipe:
Info endpoint shows incorrect version of Metronome

Summary:
The manifest file is written to dcos.jobs-x.x.x.jar file.

JIRA issues:  METRONOME-220, METRONOME-157


jobs/src/main/scala/dcos/metronome/MetronomeBuildInfo.scala
```diff
@@ -14,7 +14,7 @@ import scala.util.{ Failure, Success, Try }
   * Provides build information details regarding Metronome and Marathon at runtime.
   */
 case object MetronomeBuildInfo {
-  private val metronomeJar = "\\bmesosphere\\.metronome\\.metronome-[0-9.]+".r
```
alenkacz: does this work locally? can we write a test for this?

kensipe: it does work locally... I did test it.   I don't see how to write a test for it (it's a chicken and egg problem...  the artifact isn't built before tests and the tests would need the artifact to test).   I suppose we could mock something but I'm not sure it would really add that much value.   There are lots of ways it could break that a sim wouldn't catch (like not creating the manifest file).   What exactly are you wanting to test?   the regex match?

alenkacz: no the fact that we output meaningful version so that we don't release it in a broken form as we did before :)

project/Build.scala
```diff
@@ -170,8 +170,8 @@ object Build extends sbt.Build {
       val ScalaTest = "3.0.5"
       val ScalaCheck = "1.13.4"
       val MacWire = "2.2.5"
-      val Marathon = "1.6.322"
-      val MarathonPluginInterface = "1.6.322"
+      val Marathon = "1.6.340"
```
alenkacz: let's not update marathon in this PR and do it in a separate one. We need the 1.6.322 artifacts anyway

PR #225, opened 2018-06-27
Ken/docker param

kensipe:
<<Replace this line with your Revision Title>>

Summary:
For more details follow http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html.

JIRA issues:


docs/docs/docker.md
```diff
@@ -0,0 +1,56 @@
+---
+title: Docker Run Configurations
+---
+
+# Docker
+
+It is possible now to run jobs using docker's privilege mode and/or using docker run time parameters.  These can be used to control the [docker runtime privileges](https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities).
```
ichernetsky: nit: `docker` -> `Docker` throughout the PR.

docs/docs/docker.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Privileged Mode
+
+Running docker in privileged mode is possible by setting `privileged=true` in the docker section of the job definition.
```
ichernetsky: ... by setting `privileged` to true ... or
... by setting `privileged: true`

docs/docs/docker.md
```diff
@@ -0,0 +1,56 @@
...
+      "image": "ubuntu",
+      "privileged": true
+    }
+  }
+}
+```
+
+## Docker Parameters
+
+It is possible to set docker parameters now with job runs.   This enables the ability to change runtime capabilities of the job.   The example below removes all the default docker capabilities and adds SYSLOG.
```
ichernetsky: nit: sometimes there are two spaces between sentences, sometimes — three.

docs/index.md
```diff
@@ -15,5 +15,9 @@ Metronome documentation lives on the [DC/OS docs site](https://dcos.io/docs/1.10
 ## Contributing
 Learn how to [contribute to the Metronome project]({{ site.baseurl }}/docs/contributing.html).
 
+## Controlling Docker Runtime Capabilities
+Learn how to [change dockers runtime capabilities]({{ site.baseurl }}/docs/docker.html).
+
```
ichernetsky: One empty line should be removed to keep it consistent.

jobs/src/main/protobuf/metronome.proto
```diff
@@ -13,6 +13,14 @@ message Label {
   optional string value = 2;
 }
 
+/**
+ * A generic (key, value) pair used in various places for parameters.
+ */
+message Parameter {
+  required string key = 1;
+  required string value = 2;
```
alenkacz: shouldn't this be optional?

kensipe: marathon uses the mesos defined `Parameter` in proto which is defined as required . 
https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L2626

PR #200, opened 2018-02-15
Cmd or docker image is mandatory

alenkacz:
Summary:
Cmd or docker image should be mandatory.

JIRA issues: METRONOME-222

PR #227, opened 2018-06-27
Cmd or docker image is mandatory BP

kensipe:
Cmd or docker image is mandatory

Summary:
back port of #200 Cmd or docker image should be mandatory.

JIRA issues: METRONOME-222

### DIFF
--- a/jobs/src/main/scala/dcos/metronome/model/JobRunSpec.scala
+++ b/jobs/src/main/scala/dcos/metronome/model/JobRunSpec.scala
@@ -48,7 +48,7 @@ object JobRunSpec {
     import ViolationBuilder._
 
     override def apply(jobRunSpec: JobRunSpec): Result = {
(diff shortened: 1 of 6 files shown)


jobs/src/main/scala/dcos/metronome/model/JobRunSpec.scala
```diff
@@ -37,5 +42,21 @@ object JobRunSpec {
...
+      if (jobRunSpec.cmd.isDefined || jobRunSpec.docker.exists(d => d.image.nonEmpty))
+        Success
+      else
+        RuleViolation(jobRunSpec, JobRunSpecMessages.cmdOrDockerValidation, None)
+    }
+  }
+}
+
+object JobRunSpecMessages {
+  val cmdOrDockerValidation = "Cmd or docker image must be specified"
```
ichernetsky: nit: `cmd` or `Command` instead of `Cmd`?

alenkacz: Since this is only backport I guess we can live with this :)

PR #230, opened 2018-07-11
Gracefully handle errors during task launching

alenkacz:
See the JIRA description for full details.

JIRA issues: DCOS-39102

### DIFF
--- a/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
@@ -109,6 +109,8 @@ class JobRunExecutorActor(
 
   def addTaskToLaunchQueue(): Unit = {
     if (existsInLaunchQueue()) {
+      // we have to handle a case when actor is restarted (e.g. because of exception) and it already put something into the queue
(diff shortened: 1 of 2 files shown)


jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -107,9 +108,13 @@ class JobRunExecutorActor(
   }
 
   def addTaskToLaunchQueue(): Unit = {
-    log.info("addTaskToLaunchQueue")
-    import dcos.metronome.utils.glue.MarathonImplicits._
-    launchQueue.add(jobRun.toRunSpec, count = 1)
+    if (existsInLaunchQueue()) {
```
meichstedt: IIUC the issue was the the initial store timed out (b/c zk was slow?), the actor was restarted, tried to add to the launchQueue again, which failed because the initial store actually persisted the node.

Could you add a comment explaining why this could be the case? Not this particular scenario, but maybe something like _if the actor was restarted due to an exception, we have to check whether the jobRun was added to the launchQueue before already._

kensipe:
this solution should be back ported to `releases/0.3` and forward ported to master `0.5`

PR #233, opened 2018-07-11
Whitelist release branch for Jenkins

timcharper:
JIRA Issues: MARATHON-8313

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,7 +2,7 @@
 
 @Library('sec_ci_libs@v2-latest') _
 
-def master_branches = ["master", ] as String[]
+def master_branches = ["releases/0.4", "master", ] as String[]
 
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -2,7 +2,7 @@
 
 @Library('sec_ci_libs@v2-latest') _
 
-def master_branches = ["master", ] as String[]
+def master_branches = ["releases/0.4", "master", ] as String[]
```
kensipe: we need `releases/0.3` as well

timcharper: Honestly, I didn't need to add those other branches on Marathon because the Jenkinsfiles don't have the security plugin specified, and Community PRs use the Jenkinsfile on master.

timcharper: But not a big deal, I'll add it.

PR #237, opened 2018-08-05
Sbt 1.2.0 update

grillorafael:
This PR updates sbt version to `1.2.0`

---
**On another subject**

I saw that there was an idea of implementing an event bus for Metronome similar to Marathon Event Bus.
I am very interested in this functionality and would appreciate some guidance to develop and send a pull request.


grillorafael:
Other thing, is metronome going to be supported long term or is mesosphere planning to replace it with something else? And if it is going to be replaced, is it in the works open sourced somewhere? 

And thanks, @timcharper. I’ll try to send more contributions on this in case this project is planned be kept alive :)

kensipe:
@grillorafael thank you!   This slipped my radar but reviewing now.  
regarding metronome longevity..  we are expecting to add additional features on top of metronome (job queues)... but that is all I know for sure at the moment.    There have been talks about moving metronome functionality into marathon... but that is just talk for now.     changes would need to happen in marathon first.

regarding event bus... I haven't heard of such plans but it is a good idea.   I imagine we would do what marathon does.   And at the moment... all of the marathon code is in the metronome code as a lib... it seems like it would just need to be configured in.   I'm will to help work on a WIP PR with you if you are interested.

grillorafael:
Thanks for answering @kensipe 

I'll get more comfortable with the codebase before doing such changes so that is a work in progress now for me :)

PR #239, opened 2018-08-14
Fix SI test pipeline for Metronome

alenkacz:
We reuse as many scripts from Marathon as possible. There were many changes in those recently, we need Metronome to catch up.

JIRA issues: DCOS-40488


tests/system/Makefile
```diff
@@ -14,19 +14,24 @@ help:
...
+	pipenv sync
 
 build:
-	flake8 --count --max-line-length=120 system
+	pipenv run flake8 --count --max-line-length=120 system
 
 test:
+    # we need to make sure that the cluster is ready before we let shakedown use it
+    # the script polls the DCOS_URL until the HTTP code indicates the server is ready
+	./wait_for_cluster.sh
```
alenkacz: I was fighting with this for a long time.

Sometimes the cluster launch just ends up in a state, where not all sockets are open and then the test fails, because shakedown starts calling endpoints and that fails.

I cannot add this wait to our test setup, because some of the calls happen before the code reaches our test - it's part of shakedown booting up.

The proper way would be to fix it in shakedown/dcos-launch (?) but I already spent so much time on this that I am fine with this solution, that is working

(last three builds here https://jenkins.mesosphere.com/service/jenkins/job/system-integration-tests/job/metronome-si-dcos-open/job/av%252Ffix-si/)

jeschkies: It is odd indeed. However, let's revisit this if we ever merge Metronome into Marathon ^^.

PR #240, opened 2018-08-15
feat: expose getAllJobRuns api

grillorafael:
This pull request exposes the existing `getAllJobRuns` method under `/v1/jobs/:jobId/runs/all`


grillorafael:
Just realized that the history of runs if not stored anywhere and therefore this functionality can't be achieved the way I thought it could. Going to close it now.

PR #241, opened 2018-08-16
General deprecation fixes

grillorafael:
This pull request deals with several deprecation warnings, unused variables and so on.

It is based on #237 branch so that is why you see sbt update commit.


api/src/main/scala/dcos/metronome/api/YamlContent.scala
```diff
@@ -3,11 +3,11 @@ package api
 
 import net.jcazevedo.moultingyaml.{ YamlValue, _ }
 import play.api.http.{ ContentTypeOf, MimeTypes, Writeable }
-import play.api.mvc.{ Accepting, BodyParser, Codec, Controller }
+import play.api.mvc._
```
ichernetsky: I would prefer to have explicit imports only here.

ichernetsky:
@grillorafael, could you please resolve the merge conflict?

grillorafael:
Done

grillorafael:
All good to go from the looks of it!

It seems I can't see Metronome task list in Jira. Is there a place that I can see things that are pending to be done that I can try to do it?

ichernetsky:
@grillorafael, we've just fixed the JIRA links in the `README`. Now you should be able to fetch a list of open Metronome issues. Thank you for being involved!

grillorafael:
Thanks! I'll have a look

PR #248, opened 2018-08-22
Bump to Marathon 1.7.x

kensipe:
Bump to Marathon 1.7.x

Summary:
Bump Metronome to Marathon 1.7.50 which includes the move to the latest marathon + the refactoring of metrics.   The metrics have been integrated into metronome with the expectation that we are only using DropWizard.     The "refactoring" of metrics and the naming rules have NOT been applied yet and will come in a separate PR.   I have a strong desire to remove the metrics around Actors.    We need a conversation around this.  IMO we shouldn't have metrics / metrics names that change based on refactoring.  These should be explicitly controlled.

https://jira.mesosphere.com/browse/DCOS-40728

JIRA issues: DCOS-40728


src/main/scala/dcos/metronome/JobApplicationLoader.scala
```diff
@@ -47,6 +47,12 @@ class JobComponents(context: Context) extends BuiltInComponentsFromContext(conte
 
   override lazy val httpErrorHandler = new ErrorHandler
 
+  /*
+    Using Metrics module from marathon.   We do NOT use Kamon, so we do NOT need the 2nd param config which is used
```
ichernetsky: nit: `Marathon`

build.sbt
```diff
@@ -41,6 +41,7 @@ lazy val baseSettings = Seq(
     "-Xlint:deprecation"
   ),
   resolvers ++= Seq(
+    Resolver.JCenterRepository,
```
ichernetsky: I believe that since Kamon is not used, its exclusion rules here can be dropped too.

jobs/src/main/scala/dcos/metronome/measurement/ActorMeasurement.scala
```diff
@@ -41,14 +31,14 @@ trait ActorMeasurement { actor: Actor with ActorLogging =>
     */
   private def timePartialFunction[A, B](pf: PartialFunction[A, B]): PartialFunction[A, B] = new PartialFunction[A, B] {
     def apply(a: A): B = {
-      receiveTimer.blocking(pf.apply(a))
+      measurement.metrics.timer(s" ${actor.getClass}.receiveExceptionMeter").blocking(pf.apply(a))
```
ichernetsky: The same here.

alenkacz:
I am fine with merging this as long as we follow-up with the metrics cleanup & new names

api/src/main/scala/dcos/metronome/api/ApiModule.scala
```diff
@@ -28,6 +29,7 @@ class ApiModule(
   assets:             Assets,
   launchQueueService: LaunchQueueService,
   actorsModule:       ActorsModule,
+  metricsModule:      MetricsModule,
```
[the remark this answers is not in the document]
kensipe: yeah... macwire magic

api/src/test/scala/dcos/metronome/api/WithApplicationComponents.scala
```diff
@@ -115,6 +117,8 @@ class MockApiComponents(context: Context) extends BuiltInComponentsFromContext(c
 
   lazy val queueService: LaunchQueueService = QueueServiceFixture.simpleQueueService()
 
+  lazy val metricsModule = new DummyMetricsModule()
```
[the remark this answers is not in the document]
kensipe: yep yep...

build.sbt
kensipe: yeah... this is sad but necessary.   although metronome doesn't use kamon..  the dependency comes with using Marathon.   It does not build without this addition.

kensipe:
thanks for the fixing bump!!   
There is 1 more thing I realized while I was trying to sleep this weekend :)...  we need to change the base name for metrics from marathon to metronome... also... I would have likely missed the additional config support without @ichernetsky comment.   
I also have 1 more concern that @ichernetsky brings up.. the metricmodule support for jetty... I doubt that will work here for metronome.     One area that is strongly different from marathon for metronome is how we handle serving http requests.

kensipe:
@ichernetsky In order to handle the concerns you raised
https://jira.mesosphere.com/browse/DCOS-40963
https://jira.mesosphere.com/browse/DCOS-40964

thanks for the detail review!!!

jobs/src/test/scala/dcos/metronome/measurement/MethodMeasurementFixture.scala
```diff
@@ -1,4 +1,7 @@
 package dcos.metronome.measurement
+import mesosphere.marathon.metrics.Metrics
```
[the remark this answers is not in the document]
kensipe: good catch... and it should have used double package notation as well!

PR #249, opened 2018-08-27
Configurable Metric Reporter 

kensipe:
Configurable Metric Reporter 

Summary:
Make metric configurations and reporters available for metronome.
https://jira.mesosphere.com/browse/DCOS-40963

JIRA issues:  DCOS-40963


src/main/resources/application.conf
```diff
@@ -485,6 +485,28 @@ metronome {
...
+    statsd.port =  ${?METRONOME_STATSD_PORT}
+    statsd.transmission.interval.ms =  ${?METRONOME_STATSD_TRANSMISSION_INTERVAL_MS}
+
+    datadog.protocol =  ${?METRONOME_DATADOG_PROTOCOL}
+    datadog.host =  ${?METRONOME_DATADOG_HOST}
+    datadog.port =  ${?METRONOME_DATADOG_PORT}
+    datadog.api.key =  ${?METRONOME_DATADOG_API_KEY}
+    datadog.transmission.interval.ms =  ${?METRONOME_DATADOG_TRANSMISSION_INTERVAL_MS}
+
+    reporter.prometheus = ${?METRONOME_PROMETHEUS_ENABLED}
```
alenkacz: we don't have API endpoint for prometheus so I think we don't want this option

src/main/scala/dcos/metronome/MetronomeConfig.scala
```diff
@@ -110,6 +129,27 @@ class MetronomeConfig(configuration: Configuration) extends JobsConfig with ApiC
 
   override lazy val maxActorStartupTime: FiniteDuration = configuration.getFiniteDuration("metronome.akka.actor.startup.max").getOrElse(10.seconds)
 
+  lazy val metricsNamePrefix: String = configuration.getOptional[String]("metronome.metrics.prefix").getOrElse("metronome")
+
+  lazy val metricsPrometheusReporter: Boolean = configuration.getOptional[Boolean]("metronome.metrics.reporter.prometheus").getOrElse(false)
+  lazy val metricsStatsDReporter: Boolean = configuration.getOptional[Boolean]("metronome.metrics.reporter.statsd").getOrElse(false)
+  lazy val metricsDataDogReporter: Boolean = configuration.getOptional[Boolean]("metronome.metrics.reporter.datadog").getOrElse(false)
+
+  lazy val metricsHistogramReservoirSignificantDigits: Int = configuration.getOptional[Int]("metronome.metrics.histogram.reservoir.significant.digits").getOrElse(3)
```
alenkacz: what do you think about not duplicating the defaults here and instead relying on the defaults in Marathon? Or do you think it's best to have those duplicated and if needed change it in both places?

kensipe: yeah.. to use the marathon defaults means to not provide the configuration in metronome... I did think about conditionally providing flags / values...  I'll add in support for it... it will provide value in other ways.

src/main/resources/application.conf
kensipe: prometheus polls for metrics... I thought we set it against the `/v1/metrics` end pt...   I was expose the prometheus flag here but I don't understand it and will remove it.

and... I would rather add things when they are present... I will remove all the reporters at this point until we have support for them.

ichernetsky:
I edited the title and the description.

PR #255, opened 2018-08-29
Add metrics to the controllers

alenkacz:
JIRA issues: DCOS-40964


zen-dog:
> `"http.responses.3xx.rate"`

Didn't we change those metric names to smth. else in Marathon recently?

kensipe:
@alenkacz   @zen-dog is right... `rate` was changed to `meter`
https://github.com/mesosphere/marathon/pull/6487/files

PR #263, opened 2018-09-02
Add test for metronome config

grillorafael:
Add some tests for `MetronomeConfig` based on the issue occurred on https://github.com/dcos/metronome/commit/545acd030e34447d957ffcf01bd3fff336cb3fa9

I could not reproduce what the issue could have been so I appreciate if someone could guide into what the test should look like (i.e. what is the value in `play.server.http.port` that broke the config parsing.

I just saw that there is a suggested (Or required?) commit message pattern. I'll try to update my other PRs (and all from now on) to follow that pattern as well.

Thanks!

### DIFF
--- a/src/test/scala/dcos/metronome/MetronomeConfigTest.scala
+++ b/src/test/scala/dcos/metronome/MetronomeConfigTest.scala
@@ -1,0 +1,46 @@
+package dcos.metronome
+
+import com.typesafe.config.ConfigFactory
+import org.scalatest.{ FunSuite, GivenWhenThen, Matchers }
+import play.api.Configuration
+
+class MetronomeConfigTest extends FunSuite with Matchers with GivenWhenThen {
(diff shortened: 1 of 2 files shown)


kensipe:
@grillorafael that's exactly right... `METRONOME_PLAY_SERVER_HTTP_PORT` was set to `disabled` which has been that way for some time and others prefer the readability of that value.

grillorafael:
I’m ok with that. Just added some tests (will add more in the future) for the config to make sure at least the behavior is documented somewhere. And again by bad for breaking this.

src/test/scala/dcos/metronome/MetronomeConfigTest.scala
```diff
@@ -0,0 +1,46 @@
+package dcos.metronome
+
+import com.typesafe.config.{ Config, ConfigFactory }
```
kensipe: `Config` is not used

PR #266, opened 2018-09-10
Implement new versioning for Metronome

alenkacz:
We need unique name of every package built for Metronome. This will allow us to test them separately.

This is a proof that it works: https://s3.amazonaws.com/downloads.mesosphere.io/metronome/builds/0.5.72-8bb1367/metronome-0.5.72-8bb1367.tgz

It's the last build of this branch.

It's mostly the same stuff we did here for Marathon:

https://github.com/mesosphere/marathon/commit/e329bff2650c0c8b556d86136983ba97d10fa04d#diff-2af72f100c356273d46284f6fd1dfc08

JIRA issues: DCOS_OSS-2597


bin/install-protobuf.sh
```diff
@@ -9,6 +9,9 @@ PLATFORM=`uname`
 echo $PLATFORM
 if [ "$PLATFORM" == 'Darwin' ]; then
    DOWNLOAD_URL="https://github.com/google/protobuf/releases/download/v$VERSION/protoc-$VERSION-osx-x86_64.zip"
+else
```
alenkacz: I had to do this because I change the AMI for jenkins and that one does not have unzip

timcharper: Or maybe we don't want to change this script more than necessary :) It was just kind of weird to begin with.

kensipe: I like @timcharper recommendation

kensipe:
rebased

version.sbt
```diff
@@ -1 +0,0 @@
-version in ThisBuild := "0.5.3"
```
meln1k: Why not just delete this file then? :)

alenkacz: yep, I believe I did that...

PR #267, opened 2018-09-10
Leader needs to bounce not just a Master

kensipe:
Leader needs to bounce not just a Master

Summary:
This was originally built and run with 1 master clusters.  This now is what is intended for 1, 3 or 5 master clusters.

https://jira.mesosphere.com/browse/DCOS-40567

JIRA issues:  DCOS-40567

### DIFF
--- a/tests/system/test_root_metronome.py
+++ b/tests/system/test_root_metronome.py
@@ -265,7 +265,6 @@ def test_secret_env_var(secret_fixture):
         job_run_has_secret()
 
 
-@common.masters_exact(1)
 @shakedown.dcos_1_11
 def test_metronome_shutdown_with_no_extra_tasks():
     """ Test for METRONOME-100 regression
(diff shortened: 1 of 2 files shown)


tests/system/test_root_metronome.py
```diff
@@ -124,8 +124,9 @@ def test_disable_schedule_recovery_from_master_bounce():
         job_schedule['enabled'] = False
         client.update_schedule(job_id, 'nightly', job_schedule)
 
-        # # bounce master
-        shakedown.restart_master_node()
+        # bounce master
+        metronome_leader = shakedown.master_leader_ip()
```
alenkacz: unfortunately mesos leader != metronome leader. This does not have to be true. This is why I linked the ticket to the other ticket that is adding metronome leader info into the API...

kensipe: ah... that is what you mean :) .  great pt!   good eye!  and thanks.

tests/system/test_root_metronome.py
```diff
@@ -281,8 +281,7 @@ def test_metronome_shutdown_with_no_extra_tasks():
         # restart metronome process
```
alenkacz: we should drop this comment

kensipe: to avoid conflict and duplication I rebased this on the other PR... removing.

PR #268, opened 2018-09-10
Turning on DC/OS 1.11 SI test

kensipe:
Turning on DC/OS 1.11 SI test

Summary:
Providing context for skipping test on cluster earlier than 1.8.   Turning on test which should have been on already.

JIRA issues:

### DIFF
--- a/tests/system/test_root_metronome.py
+++ b/tests/system/test_root_metronome.py
@@ -265,7 +265,6 @@ def test_secret_env_var(secret_fixture):
         job_run_has_secret()
 
 
-@common.masters_exact(1)
 @shakedown.dcos_1_11
 def test_metronome_shutdown_with_no_extra_tasks():
     """ Test for METRONOME-100 regression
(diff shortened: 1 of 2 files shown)


tests/system/test_root_metronome.py
```diff
@@ -265,7 +266,7 @@ def job_run_has_secret():
 
 
 @common.masters_exact(1)
-@pytest.mark.skip(reason="we need to wait until METRONOME-100 gets to testing/master")
```
alenkacz: I guess you verified on CCM cluster with at least 10 runs that this does not fail, right? I believe the reason why this is also commented is that it fetches metronome leader and when we bounce a leader in a different test, that might not be accurate. I still think we need that leader endpoint

kensipe: @alenkacz I don't understand... this test uses the leader
`metronome_leader = shakedown.master_leader_ip()`

PR #275, opened 2018-10-04
Crash when exception during loading occurs

alenkacz:
Right now we were swallowing up these exceptions.

JIRA issues: DCOS_OSS-4239

### DIFF
--- a/src/main/scala/dcos/metronome/JobApplicationLoader.scala
+++ b/src/main/scala/dcos/metronome/JobApplicationLoader.scala
@@ -4,24 +4,31 @@ import java.time.Clock
 
 import controllers.AssetsComponents
 import com.softwaremill.macwire._
+import com.typesafe.scalalogging.StrictLogging
 import dcos.metronome.api.v1.LeaderProxyFilter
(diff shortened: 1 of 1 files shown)


src/main/scala/dcos/metronome/JobApplicationLoader.scala
```diff
@@ -4,32 +4,43 @@ import java.time.Clock
...
+
   def load(context: Context): Application = {
     val jobComponents = new JobComponents(context)
 
     jobComponents.metricsModule.start(jobComponents.actorSystem)
 
     Future {
       jobComponents.schedulerService.run()
-    }(scala.concurrent.ExecutionContext.global)
+    }(scala.concurrent.ExecutionContext.global).onComplete {
```
timcharper: Not related to your pull request, but I just realized how terrible this is. Were blocking one of the global execution contexts threads for the entire life of metronome, just so the scheduler service can have something to block?

alenkacz: haha, good point, that's not really nice...

PR #282, opened 2018-10-16
No timeout for jenkins job

alenkacz:
Summary:
Right now we have two timeouts:
- one on the jenkins job level
- one on the pipeline level https://github.com/dcos/metronome/blob/master/ci/si_pipeline.sh#L79

The internal pipeline timeout is enough.

JIRA issues: MARATHON-8479


zen-dog:
Why 2 timeouts at all? In marathon scripts we got rid of Jenkins level timeout - I'd suggest the same here.

PR #287, opened 2018-11-08
Replace EQ operator with IS

timcharper:
We replace the EQ operator with IS to better match Marathon's constraint API. Change is done in a backwards compatible way such that:

* Posting an EQ constraint gets converted to an IS constraint, automatically
* Upgrading from a prior version of metronome with a JobSpec transparently converts to IS
* Valueless IS makes no sense (as does LIKE and UNLIKE) and continues to have no effect, as valueless EQ did.

JIRA Issues: DCOS_OSS-4464

### DIFF
--- a/api/src/main/scala/dcos/metronome/api/v1/models/package.scala
+++ b/api/src/main/scala/dcos/metronome/api/v1/models/package.scala
@@ -106,7 +106,7 @@ package object models {
   }
 
   implicit lazy val OperatorFormat: Format[Operator] = new Format[Operator] {
-    override def writes(o: Operator): JsValue = JsString(Operator.name(o))
+    override def writes(o: Operator): JsValue = JsString(o.name)
     override def reads(json: JsValue): JsResult[Operator] = json match {
       case JsString(Operator(value)) => JsSuccess(value)
       case invalid => JsError(s"'$invalid' is not a valid operator. " +
--- a/jobs/src/main/scala/dcos/metronome/model/PlacementSpec.scala
+++ b/jobs/src/main/scala/dcos/metronome/model/PlacementSpec.scala
@@ -9,19 +9,18 @@ object PlacementSpec {
 
 case class ConstraintSpec(attribute: String, operator: Operator, value: Option[String])
 
-sealed trait Operator
+sealed trait Operator { val name: String }
 object Operator {
-  case object Eq extends Operator
-  case object Like extends Operator
-  case object Unlike extends Operator
+  case object Is extends Operator { val name = "IS" }
+  case object Like extends Operator { val name = "LIKE" }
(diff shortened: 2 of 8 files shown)


api/src/main/scala/dcos/metronome/api/v1/models/package.scala
```diff
@@ -106,7 +106,7 @@ package object models {
   }
 
   implicit lazy val OperatorFormat: Format[Operator] = new Format[Operator] {
-    override def writes(o: Operator): JsValue = JsString(Operator.name(o))
+    override def writes(o: Operator): JsValue = JsString(o.name)
```
kensipe: I really dislike when we use variable names like `o`... I suppose in this context it is so small as to be readable but is general frustrating.

jobs/src/main/scala/dcos/metronome/model/PlacementSpec.scala
```diff
@@ -9,19 +9,18 @@ object PlacementSpec {
...
-    "UNLIKE" -> Unlike)
-  val operatorNames: Map[Operator, String] = names.map{ case (a, b) => (b, a) }
-
-  def name(operator: Operator): String = operatorNames(operator)
+  val all = Seq(Is, Like, Unlike)
+  /**
+    * For backwards compatibility, we map EQ to Is
+    */
+  val names: Map[String, Operator] =
+    Map(all.map { op => op.name -> op }: _*) + ("EQ" -> Is)
```
kensipe: nice

jobs/src/main/scala/dcos/metronome/utils/glue/MarathonImplicits.scala
```diff
@@ -39,18 +39,28 @@ object MarathonImplicits {
...
-      val marathonOperator = spec.operator match {
-        case Operator.Eq     => marathon.Protos.Constraint.Operator.CLUSTER
-        case Operator.Like   => marathon.Protos.Constraint.Operator.LIKE
-        case Operator.Unlike => marathon.Protos.Constraint.Operator.UNLIKE
-      }
+    def toProto: Option[marathon.Protos.Constraint] = {
+      /**
+        * Unfortunately, Metronome has always allowed valueless constraint operators, but they never had any effect.
+        *
+        * The Eq operator was replaced with Is in of favor consistency. Eq mapped to CLUSTER, and valueless cluster in
```
kensipe: strongly with parsing "with Is in of favor consistency" .  I think you mean "in favor of consistency"

timcharper: Hahahah, good catch. I can haz grammar.

timcharper:
@kensipe I was waiting on one more review for these :)

PR #290, opened 2018-11-12
 Map UCR Changes to App Def

meln1k:
Summary: added support for UCR when we create an app definition. Zookeeper serialization and http API marshalling are also updated.

Jira Issues: DCOS_OSS-4437

### DIFF
--- a/jobs/src/main/scala/dcos/metronome/model/Container.scala
+++ b/jobs/src/main/scala/dcos/metronome/model/Container.scala
@@ -27,5 +27,5 @@ object ImageSpec {
 
 case class UcrSpec(
   image:      ImageSpec,
-  privileged: Boolean   = false)
+  privileged: Boolean   = false) extends Container
 
--- a/jobs/src/main/scala/dcos/metronome/model/Container.scala
+++ b/jobs/src/main/scala/dcos/metronome/model/Container.scala
@@ -15,3 +15,17 @@ case class DockerSpec(
 object DockerSpec {
   val DefaultParameters = Seq.empty[Parameter]
 }
+
+case class ImageSpec(
+  id:        String,
+  kind:      String  = ImageSpec.DefaultKind,
+  forcePull: Boolean = false)
+
+object ImageSpec {
+  val DefaultKind = "docker"
+}
+
+case class UcrSpec(
+  image:      ImageSpec,
+  privileged: Boolean   = false)
+
--- a/jobs/src/main/scala/dcos/metronome/model/JobRunSpec.scala
+++ b/jobs/src/main/scala/dcos/metronome/model/JobRunSpec.scala
@@ -68,7 +68,7 @@ object JobRunSpec {
 }
 
 object JobRunSpecMessages {
-  val cmdOrDockerValidation = "Cmd, docker or ucr image must be specified"
+  val cmdOrDockerValidation = "Cmd, Docker or UCR image must be specified"
   def secretsValidation(envVarSecretsName: Set[String], providedSecretsNames: Set[String]) = {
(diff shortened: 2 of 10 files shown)


jobs/src/main/scala/dcos/metronome/utils/glue/MarathonImplicits.scala
```diff
@@ -72,7 +72,8 @@ object MarathonImplicits {
 
   implicit class JobSpecToContainer(val jobSpec: JobSpec) extends AnyVal {
     def toContainer: Option[Container] = {
-      jobSpec.run.docker match {
+      require(!(jobSpec.run.docker.nonEmpty && jobSpec.run.ucr.nonEmpty), "docker and ucr can't be present both")
```
wavesoft: I like the idea of failing loud and proud here, though I don't like the fact that we actually have to _do_ this kind of check. Ideally I would like to see a `DockerJobSpec` and a `UCRJobSpec`, bot deriving from the same `JobSpec`.

Nonetheless, this is working right now and we could add this on tech debt on a next revision, so I am not going to block the PR.

jobs/src/main/scala/dcos/metronome/model/Container.scala
```diff
@@ -15,3 +15,17 @@ case class DockerSpec(
...
   val DefaultParameters = Seq.empty[Parameter]
 }
+
+case class ImageSpec(
+  id:        String,
+  kind:      String  = ImageSpec.DefaultKind,
+  forcePull: Boolean = false)
+
+object ImageSpec {
+  val DefaultKind = "docker"
```
jeschkies: Nit: `DefaultKind` seems an odd name when it is `docker`. Why not `ImageSpec.Docker`?

jobs/src/main/scala/dcos/metronome/model/JobRunSpec.scala
```diff
@@ -58,17 +58,19 @@ object JobRunSpec {
...
+      check(!(jobRunSpec.docker.nonEmpty & jobRunSpec.ucr.nonEmpty), JobRunSpecMessages.onlyDockerOrUcr)
 
       violations.headOption.getOrElse(Success)
     }
   }
 }
 
 object JobRunSpecMessages {
-  val cmdOrDockerValidation = "Cmd or docker image must be specified"
+  val cmdOrDockerValidation = "Cmd, docker or ucr image must be specified"
```
kensipe: we should have consistency with line 75 below... Docker (not docker) and UCR.   Also this message could be more clear IMO.

jobs/src/main/scala/dcos/metronome/utils/glue/MarathonImplicits.scala
```diff
@@ -82,6 +83,13 @@ object MarathonImplicits {
         case _ => None
       }
 
+      val maybeUcr = jobSpec.run.ucr.map { ucrSpec =>
+        Container.MesosDocker(
+          image = ucrSpec.image.id,
+          forcePullImage = ucrSpec.image.forcePull) // TODO: pass privileged once marathon will support it
+      }
+
+      maybeDocker.orElse(maybeUcr)
```
kensipe: does this allow for UCR to be None?   it is possible to have UCR and Docker be None

jobs/src/main/scala/dcos/metronome/model/Container.scala
kensipe: The way it is written is consistent with the rest of the metronome code base where defaults are defined and easily understood.  It doesn't seem necessary here as clearly line 21 would indicate that it is the default.    
the file `models/pacakge.scala` shows uses of defaults for reading from JSON.   It would be clear to change to `ImageSpec.Docker`... however it would be inconsistent with everything else in that file.

PR #291, opened 2018-11-12
Docs for UCR support for Jobs

meln1k:
Summary: added a documentation page which provides an example of UCR usage.

Jira Issues: DCOS_OSS-4439

### DIFF
--- a/docs/docs/ucr.md
+++ b/docs/docs/ucr.md
@@ -1,0 +1,29 @@
+---
+title: Universal Container Runtime (UCR) Run Configurations
+---
+
+# UCR
+
+It is now possible to run jobs with Docker containers using [Universal Container Runtime](http://mesos.apache.org/documentation/latest/container-image/).
+
+To enable UCR, the run configuration needs to contain the `ucr` section:
+
+```
+{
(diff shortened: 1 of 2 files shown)


docs/docs/ucr.md
```diff
@@ -0,0 +1,29 @@
...
+      "image":  {
+        "id": "ubuntu",
+        "forcePull": true
+      }
+    }
+  }
+}
+```
+
+Please note that UCR uses a differnt syntax when defining docker image section. Also, specifying both `docker` and `ucr` will lead to a validation error.
```
wavesoft: I would move this before the example and express it like so:

```markdown
**Important:** Note that UCR could support various image types, therefore the `ucr.image` property is 
an _object_ instead of a _string_ as with `docker`. Specifying just `"image": { "id": "ubuntu" }` will 
default to `docker` image type.
```

And then, _after_ the example, add:

```markdown
Note that specifying both `ucr` and `docker` is invalid and will lead to a validation error.
```

meln1k: I like it! 👍

PR #295, opened 2018-11-16
Job Schema to App Def Mapping for File Based Secrets

meln1k:
Summary: added support for File Based Secrets when we create an app definition. Zookeeper serialization and http API marshalling are also updated.

Jira Issues: DCOS_OSS-4448

### DIFF
--- a/jobs/src/main/scala/dcos/metronome/model/JobRunSpec.scala
+++ b/jobs/src/main/scala/dcos/metronome/model/JobRunSpec.scala
@@ -55,12 +55,21 @@ object JobRunSpec {
           violations += RuleViolation(jobRunSpec, errorMessage)
         }
       }
-      val envVarDefinedSecretNames = jobRunSpec.env.values.collect { case EnvVarSecret(secretName) => secretName }.toSet
+      val definedSecretNames =
+        jobRunSpec.env.values.collect { case EnvVarSecret(secretName) => secretName }.toSet ++
+          jobRunSpec.volumes.collect { case SecretVolume(_, secret) => secret }.toSet
       val providedSecretNames = jobRunSpec.secrets.keySet
 
       check(jobRunSpec.cmd.isDefined || jobRunSpec.docker.exists(d => d.image.nonEmpty) || jobRunSpec.ucr.nonEmpty, JobRunSpecMessages.cmdOrDockerValidation)
-      check(envVarDefinedSecretNames == providedSecretNames, JobRunSpecMessages.secretsValidation(envVarDefinedSecretNames, providedSecretNames))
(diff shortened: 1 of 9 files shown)


jobs/src/test/scala/dcos/metronome/repository/impl/kv/marshaller/JobSpecMarshallerTest.scala
```diff
@@ -114,38 +120,4 @@ class JobSpecMarshallerTest extends FunSuite with Matchers with PropertyChecks {
     val converted = Seq(Constraint.newBuilder().setAttribute("field").setOperator(Constraint.Operator.EQ).setValue("value").build).toModel
     converted.head.operator.name shouldBe "IS"
   }
-
-  class Fixture {
```
meln1k: This is not used anywhere, so I deleted it as dead code.

jobs/src/main/scala/dcos/metronome/model/JobRunSpec.scala
```diff
@@ -55,13 +55,22 @@ object JobRunSpec {
...
       }
-      val envVarDefinedSecretNames = jobRunSpec.env.values.collect { case EnvVarSecret(secretName) => secretName }.toSet
+      val definedSecretNames =
+        jobRunSpec.env.values.collect { case EnvVarSecret(secretName) => secretName }.toSet ++
+          jobRunSpec.volumes.collect { case SecretVolume(_, secret) => secret }.toSet
       val providedSecretNames = jobRunSpec.secrets.keySet
 
       check(jobRunSpec.cmd.isDefined || jobRunSpec.docker.exists(d => d.image.nonEmpty) || jobRunSpec.ucr.nonEmpty, JobRunSpecMessages.cmdOrDockerValidation)
-      check(envVarDefinedSecretNames == providedSecretNames, JobRunSpecMessages.secretsValidation(envVarDefinedSecretNames, providedSecretNames))
+      check(definedSecretNames == providedSecretNames, JobRunSpecMessages.secretsValidation(definedSecretNames, providedSecretNames))
```
kensipe: I'm not sure that the `definedSecretNames` needs to == `definedSecretNames`.   What is required is that a `definedSecretNames` is in the set of providedSecretNames.
granted... I don't have a use case for providing a secret which isn't used... however I could see someone debugging / troubleshooting and removing the use of the secret and being frustrated with needing to also remove the provided secret name as well.    
This just seems to be more strict than necessary

timcharper: Queue unused imports discussion 😆 🤣 

There's utility in it being strict and in this case I don't see much drawback to being strict. I'd vote to keep the strictness.

(And, also, I'd vote against making it more strict than it was, but that doesn't seem to be the case here)

PR #299, opened 2018-11-19
SI Tests for File Based Secrets

meln1k:
Summary: added a SI test similar to the env secrets test but this time for File Based Secrets

Jira issues: DCOS_OSS-4449


kensipe:
woot!!

PR #300, opened 2018-11-19
Job Schema Design for GPU

kensipe:
Job Schema Design for GPU

Summary:
Schema needed for landing on GPU in metronome

JIRA issues:   DCOS_OSS-4441


api/src/main/resources/public/api/v0/schema/jobspec_v0.schema.json
```diff
@@ -74,6 +74,11 @@
           "description": "The number of CPU shares this job needs per instance. This number does not have to be integer, but can be a fraction.",
           "minimum": 0.01
         },
+        "gpus": {
+          "type": "number",
+          "description": "The number of GPU shares this job needs per instance. This number does not have to be integer, but can be a fraction.",
```
alenkacz: I think GPUs can't be a fraction, it has to be an integer...

PR #301, opened 2018-11-20
Job Schema to App Def Mapping for GPUs

meln1k:
Summary: added support for GPUs when we create an app definition.
Zookeeper serialization and http API marshalling are also updated.

Jira Issues: DCOS_OSS-4443

### DIFF
--- a/jobs/src/main/scala/dcos/metronome/model/JobRunSpec.scala
+++ b/jobs/src/main/scala/dcos/metronome/model/JobRunSpec.scala
@@ -11,6 +11,7 @@ case class JobRunSpec(
   cpus:                       Double                           = JobRunSpec.DefaultCpus,
   mem:                        Double                           = JobRunSpec.DefaultMem,
   disk:                       Double                           = JobRunSpec.DefaultDisk,
+  gpus:                       Int                              = JobRunSpec.DefaultGpus,
   cmd:                        Option[String]                   = JobRunSpec.DefaultCmd,
   args:                       Option[Seq[String]]              = JobRunSpec.DefaultArgs,
   user:                       Option[String]                   = JobRunSpec.DefaultUser,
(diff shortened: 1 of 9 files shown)


jobs/src/main/scala/dcos/metronome/model/JobRunSpec.scala
```diff
@@ -71,6 +73,8 @@ object JobRunSpec {
       def noSecretVolumesExists: Boolean = jobRunSpec.volumes.forall(v => !isSecretVolume(v))
       check(noSecretVolumesExists || jobRunSpec.ucr.isDefined, JobRunSpecMessages.fileBasedSecretsAreUcrOnly)
 
+      check(jobRunSpec.gpus == 0 || jobRunSpec.ucr.isDefined, JobRunSpecMessages.gpusAreUcrOnly)
```
kensipe: this check seems off...    we shouldn't short circuit with `||`... if gpus are > 0.... ucr.isDefined isn't evaluated because it wouldn't matter whats value is for the value to be true.    It also allows for gpus > 0 and ucr to not be defined.   We want an ADD here.

jobs/src/main/scala/dcos/metronome/model/JobRunSpec.scala
```diff
@@ -71,6 +73,8 @@ object JobRunSpec {
       def noSecretVolumesExists: Boolean = jobRunSpec.volumes.forall(v => !isSecretVolume(v))
       check(noSecretVolumesExists || jobRunSpec.ucr.isDefined, JobRunSpecMessages.fileBasedSecretsAreUcrOnly)
```
kensipe: I don't like this either... now that I'm reviewing again...

kensipe: can my mind... this one is perfect... it is a great short circuit... if noSecretVolumes... nothing left to check.

jobs/src/main/scala/dcos/metronome/model/JobRunSpec.scala
kensipe: haha... ok.. I'm good with it.   there is a lot of negations in this code... it is challenging to read and mentally validate...

PR #303, opened 2018-11-20
Fixes for secrets documentation

meln1k:
Summary: Replaced flags with env vars.

Jira Issues: DCOS_OSS-4450


kensipe:
@meln1k there is some thought on defaulting secrets as to always on... what say you?
if secrets is enabled AND there is no secret plugin secrets are ignored... which seems fine.   It would simplify things with no ill effect.

PR #310, opened 2018-12-04
Do not dump whole config when starting metronome

alenkacz:


alenkacz:
This needs to be backported!

PR #311, opened 2018-12-04
Resolving timeout issues associated with Master bounce

kensipe:
Resolving timeout issues associated with Master bounce

Summary:
decrease timeout for each test.  remove reference to missing code from copy and paste mistake

total test timeout is changed from 2 hours to 1.5 hours.  This is to ensure that the cluster is still up after the timeout and before the cluster is destroyed... so we can get diag bundles.

JIRA issues:  DCOS_OSS-4536

### DIFF
--- a/tests/system/Makefile
+++ b/tests/system/Makefile
@@ -28,7 +28,7 @@ test:
       --stdout all \
       --stdout-inline \
       --ssl-no-verify \
-      --timeout 360000 \
+      --timeout 1800 \
       --pytest-option "--junitxml=../../shakedown.xml" \
       --pytest-option --verbose \
       --pytest-option --full-trace \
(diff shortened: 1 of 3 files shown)


tests/system/Makefile
```diff
@@ -28,7 +28,7 @@ test:
       --stdout all \
       --stdout-inline \
       --ssl-no-verify \
-      --timeout 360000 \
+      --timeout 1800 \
```
alenkacz: I don't understand why did we made the timeout smaller, what is this fixing? Thanks

kensipe:
we don't have the info we need to solve the underlying issue...   we need a solution to:
1) having appropriate timeouts
2) not calling code that doesn't exist

with these changes we should be able to get the diagnostic bundle... which will provide appropriate insight in the future

tests/system/Makefile
kensipe: yes it is fixing something... details in the jira... but the outer timeout using the unix `timeout` was 2 hrs... first the cluster expires in 2 hrs... if we wait that long we can't get a diagnostic bundle.  second that is all or nothing.   This change ensures that each individual test does not go longer than 30 mins.   If there is 1 long running test that is causing these issues (which is what is suspected)... then that test will not run for 1-2 hrs... it will be killed in 30 mins and we move on.   
We are also changing the timeout because whoever originally assigned the time likely didn't understand that this isn't in milliseconds... it is seconds.    Do we really want a timeout that is 100 hours?

PR #314, opened 2018-12-07
Making FBS work from v1 and working SI

kensipe:
Making FBS work from v1 and working SI

Summary:
Testing FBS from v1 was failing.   It required a `/` for a containerpath. which breaks secrets for some reason.    I was able to test this from v0 which doesn't have this restriction.   The SI tests were broken as well and have been modified to be less flaky.

JIRA issues:

### DIFF
--- a/api/src/main/resources/public/api/v1/schema/jobspec.schema.json
+++ b/api/src/main/resources/public/api/v1/schema/jobspec.schema.json
@@ -250,8 +250,7 @@
               "containerPath": {
                 "type": "string",
                 "description": "The path of the volume in the container",
-                "minLength": 1,
-                "pattern": "^/[^/].*$"
+                "minLength": 1
               },
               "hostPath": {
                 "type": "string",
--- a/tests/system/common.py
+++ b/tests/system/common.py
@@ -59,6 +59,7 @@ def job_with_file_based_secret(
         id='pikachu-fbs',
         cmd='cat $MESOS_SANDBOX/secret-file > $MESOS_SANDBOX/fbs-secret; sleep 30',
         secret_name='secret_name'):
+    # secret container path can not have '/' prefix for secret, otherwise needs it
     return {
         'id': id,
         'description': 'electrifying rodent',
--- a/tests/system/common.py
+++ b/tests/system/common.py
@@ -57,7 +57,7 @@ def job_with_secrets(id='pikachu',
 
 def job_with_file_based_secret(
         id='pikachu-fbs',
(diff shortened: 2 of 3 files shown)


api/src/main/resources/public/api/v1/schema/jobspec.schema.json
```diff
@@ -250,8 +250,7 @@
               "containerPath": {
                 "type": "string",
                 "description": "The path of the volume in the container",
-                "minLength": 1,
-                "pattern": "^/[^/].*$"
+                "minLength": 1
```
kensipe: v0 did NOT require this and we can NOT require it for v1.   The pattern is not allowed for FBS.

tests/system/common.py
```diff
@@ -69,20 +70,20 @@ def job_with_file_based_secret(
...
-                    "containerPath": "/secret-file",
+                    "containerPath": "secret-file",
                     "secret": "secret1"
                 }
             ],
+            'ucr': {
+                "image": {
+                    "id": "busybox"
+                }
+            },
```
kensipe: this was just broken and was in the wrong location based on schema needs.

tests/system/common.py
```diff
@@ -69,20 +70,20 @@ def job_with_file_based_secret(
             'disk': 0,
             "volumes": [
                 {
-                    "containerPath": "/secret-file",
+                    "containerPath": "secret-file",
```
kensipe: '`' breaks FBS

tests/system/common.py
```diff
@@ -57,8 +57,9 @@ def job_with_secrets(id='pikachu',
 
 def job_with_file_based_secret(
         id='pikachu-fbs',
-        cmd='cat $MESOS_SANDBOX/secret-file > $MESOS_SANDBOX/fbs-secret; sleep 5',
+        cmd='cat $MESOS_SANDBOX/secret-file > $MESOS_SANDBOX/fbs-secret; sleep 30',
```
kensipe: the sleep time is not important... the job and all running tasks will be destroyed as soon as the test passes.

api/src/main/resources/public/api/v1/schema/jobspec.schema.json
timcharper: Oh, yeah, relative paths are totes supposed to be supported, because they go in the Mesos sandbox, right?

tests/system/common.py
timcharper: `/` breaks FBS?

kensipe: yes

PR #316, opened 2018-12-18
Test Mesos Not AWS for Master Bounce

kensipe:
Test Mesos Not AWS for Master Bounce

Summary:
do not restart node, just restart the master agent. also lets provide cluster info for test which includes version.

JIRA issues:

### DIFF
--- a/tests/system/test_root_metronome.py
+++ b/tests/system/test_root_metronome.py
@@ -126,7 +126,7 @@ def test_disable_schedule_recovery_from_master_bounce():
         job_schedule['enabled'] = False
         client.update_schedule(job_id, 'nightly', job_schedule)
 
-        # bounce metronome master
+        # bounce mesos master
         shakedown.run_command_on_leader('sudo systemctl restart dcos-mesos-master')
         common.wait_for_cosmos()
         common.wait_for_metronome()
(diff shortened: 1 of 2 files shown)


tests/system/test_root_metronome.py
```diff
@@ -127,7 +127,7 @@ def test_disable_schedule_recovery_from_master_bounce():
         client.update_schedule(job_id, 'nightly', job_schedule)
 
         # bounce metronome master
-        common.run_command_on_metronome_leader('sudo /sbin/shutdown -r now')
+        shakedown.run_command_on_leader('sudo systemctl restart dcos-mesos-master')
```
timcharper: This comment says "bounce metronome master" but the command says dcos-mesos-master. Comment needs to be updated?

tests/system/test_root_metronome.py
```diff
@@ -327,6 +327,7 @@ def test_metronome_shutdown_with_no_extra_tasks():
 def setup_module(module):
     common.wait_for_metronome()
     common.wait_for_cosmos()
+    common.cluster_info()
```
timcharper: is this just to block until ready or something? It seems like we're calling cluster_info and discarding the value, and that's not clear why we'd do that.

PR #323, opened 2019-03-19
Metronome must have STATSD metrics configurable

kensipe:
Summary:
Adding statd configuration for metronome which is passed through to the underlying marathon

JIRA issues:   DCOS_OSS-4936


kensipe:
Makes you wonder... if an application has metrics and it can't be collected does anyone hear it?

PR #325, opened 2019-04-01
Fixing the ability to have a failed job have a retry

kensipe:
Summary:
Previous bug fix introduced this bug which prevented metronome from restarting a failed job run.

JIRA issues:  DCOS_OSS-4636


alenkacz:
Also we should backport this!

kensipe:
@alenkacz regarding tests... the question is why did `RestartPolicy is handled correctly` test pass green with the previous impl.    looking into it.

alenkacz:
@kensipe ooh, we have tests already, hmmm ... it would be good to understand then. Good that you're looking into it, thanks 👍

PR #326, opened 2019-04-02
Add metric for uptime

alenkacz:
We don't inherit this from Marathon because in Marathon we measure this in `GroupManagerModule`.

JIRA issues: DCOS-50467


src/main/scala/dcos/metronome/JobApplicationLoader.scala
```diff
@@ -36,6 +37,11 @@ class JobApplicationLoader extends ApplicationLoader with StrictLogging {
 
     jobComponents.metricsModule.start(jobComponents.actorSystem)
 
+    val startedAt = System.currentTimeMillis()
```
zen-dog: Maybe move `startedAt` into class private val in case `load()` is called multiple times?

alenkacz: I think it should not be called multiple times, we would have more problems if it was but done ✅

PR #331, opened 2019-04-05
Improve secrets validation to only point out unprovided secrets

meichstedt:
Summary:
This commit adjusts secrets validation to provide helpful error messages when secrets are defined but not referenced, or referenced but not defined.

JIRA issues: DCOS_OSS-5019

### DIFF
--- a/api/src/test/scala/dcos/metronome/api/v1/controllers/JobSpecControllerTest.scala
+++ b/api/src/test/scala/dcos/metronome/api/v1/controllers/JobSpecControllerTest.scala
@@ -294,6 +294,7 @@ class JobSpecControllerTest extends PlaySpec with OneAppPerTestWithComponents[Mo
       Then("A validation error is returned")
       status(response) mustBe UNPROCESSABLE_ENTITY
       contentType(response) mustBe Some("application/json")
+      contentAsString(response).contains("secretId") mustBe true
     }
 
     "indicate a problem when creating a job which contains fields both for secret and normal volume" in {
--- a/api/src/test/scala/dcos/metronome/api/v1/controllers/JobSpecControllerTest.scala
+++ b/api/src/test/scala/dcos/metronome/api/v1/controllers/JobSpecControllerTest.scala
@@ -285,16 +285,15 @@ class JobSpecControllerTest extends PlaySpec with OneAppPerTestWithComponents[Mo
       contentType(response) mustBe Some("application/json")
     }
 
-    "indicate a problem when creating a job without a secret name" in {
+    "indicate no problem when creating a job with unused secrets" in {
       Given("No job")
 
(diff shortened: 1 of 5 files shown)


api/src/test/scala/dcos/metronome/api/v1/controllers/JobSpecControllerTest.scala
```diff
@@ -285,16 +285,15 @@ class JobSpecControllerTest extends PlaySpec with OneAppPerTestWithComponents[Mo
       contentType(response) mustBe Some("application/json")
     }
 
-    "indicate a problem when creating a job without a secret name" in {
+    "indicate no problem when creating a job with unused secrets" in {
```
alenkacz: it took me time to figure out why we needed this change but 👍 

Btw. it is a bit controversial because having extra secret might mean that someone forgot to actually use it which he/she will figure out too late, in the runtime...

meichstedt: I've changed it (and the JIRA description) to fail in both scenarios, with a meaningful message.

jobs/src/test/scala/dcos/metronome/model/JobRunSpecValidatorTest.scala
```diff
@@ -0,0 +1,61 @@
...
+    Given("a JobRunSpec with undefined, but referenced secrets")
+    val jobRunSpec = JobRunSpec(
+      cmd = Some("sleep 1000"),
+      volumes = Seq(SecretVolume("path", "undefined-vol-secret")))
+
+    When("The spec is validated")
+    val result = JobRunSpec.validJobRunSpec(jobRunSpec)
+
+    Then("a validation error is returned")
+    result.isFailure shouldBe true
```
zen-dog: Wdyt about checking that the undefined secret is mentioned in the failure message here and below?

zen-dog: Probably not really necessary.

meichstedt: Yeah – We have some helper code in Marathon that parses those, but I didn't want to add this. No test in here really verifies the message matches the expectations.

Only checking for `contains` is probably easy enough, and has _some_ value.

zen-dog: Yep, our Marathon "helpers" just check for `string.contains()` :D

meichstedt: Yep, I added these as well. thanks for the suggestion :)

PR #333, opened 2019-04-08
Fix marathon version reported in info endpoint

alenkacz:
The marathon version was incorrectly reported, it was always the default of 1.7.0-SNAPSHOT. I believe that we cannot use the same regex as we use in marathon here thus I copypasted the build info class and changed the regex.

### DIFF
--- a/jobs/src/main/scala/dcos/metronome/MarathonBuildInfo.scala
+++ b/jobs/src/main/scala/dcos/metronome/MarathonBuildInfo.scala
@@ -1,0 +1,64 @@
+package dcos.metronome
+
+import java.util.jar.{ Attributes, Manifest }
+
+import mesosphere.marathon.SemVer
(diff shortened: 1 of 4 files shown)


jobs/src/main/scala/dcos/metronome/MarathonBuildInfo.scala
```diff
@@ -0,0 +1,64 @@
...
+
+import mesosphere.marathon.SemVer
+import mesosphere.marathon.io.IO
+import mesosphere.marathon.stream.Implicits._
+
+import scala.Predef._
+import scala.util.control.NonFatal
+
+case object MarathonBuildInfo {
+  private val marathonJar = """\/mesosphere\/marathon\/marathon_2.12\/[0-9.]+""".r
```
alenkacz: this is the only change

PR #334, opened 2019-04-09
Remove timer.seconds as that is already added

alenkacz:
There is convention that already does that if we use these interfaces/implementation we use. See https://github.com/mesosphere/marathon/blob/db1da90ae6515f55faf59b110f18851c3ad2134a/src/main/scala/mesosphere/marathon/metrics/current/DropwizardMetrics.scala#L104

### DIFF
--- a/api/src/main/scala/dcos/metronome/api/Authorization.scala
+++ b/api/src/main/scala/dcos/metronome/api/Authorization.scala
@@ -86,7 +86,7 @@ abstract class Authorization(
   private[this] val http4XX = metrics.counter("http.responses.4xx.rate")
   private[this] val http5XX = metrics.counter("http.responses.5xx.rate")
   private[this] val apiErrors = metrics.counter("http.responses.errors.rate")
-  private[this] val requestDurationMetric = metrics.timer("http.requests.duration.timer.seconds")
+  private[this] val requestDurationMetric = metrics.timer("http.requests.duration")
   private[this] val requestSizeMetric = metrics.counter("http.requests.size")
   private[this] val responseSizeMetric = metrics.counter("http.responses.size")
 


api/src/main/scala/dcos/metronome/api/Authorization.scala
```diff
@@ -86,9 +87,9 @@ abstract class Authorization(
   private[this] val http4XX = metrics.counter("http.responses.4xx.rate")
   private[this] val http5XX = metrics.counter("http.responses.5xx.rate")
   private[this] val apiErrors = metrics.counter("http.responses.errors.rate")
-  private[this] val requestDurationMetric = metrics.timer("http.requests.duration.timer.seconds")
-  private[this] val requestSizeMetric = metrics.counter("http.requests.size")
-  private[this] val responseSizeMetric = metrics.counter("http.responses.size")
+  private[this] val requestDurationMetric = metrics.timer("http.requests.duration")
+  private[this] val requestSizeMetric = metrics.counter("http.requests.size", UnitOfMeasurement.Memory)
```
alenkacz: this will allow metric to append `.bytes` to the name

api/src/main/scala/dcos/metronome/api/Authorization.scala
```diff
@@ -80,15 +81,15 @@ abstract class Authorization(
     }
   }
 
-  private[this] val http1XX = metrics.counter("http.responses.1xx.rate")
-  private[this] val http2XX = metrics.counter("http.responses.2xx.rate")
-  private[this] val http3XX = metrics.counter("http.responses.3xx.rate")
-  private[this] val http4XX = metrics.counter("http.responses.4xx.rate")
-  private[this] val http5XX = metrics.counter("http.responses.5xx.rate")
+  private[this] val http1XX = metrics.meter("http.responses.1xx.rate")
```
alenkacz: I think this was bug even in marathon. We now use meter in marathon as well

PR #339, opened 2019-06-11
Update status even when active

alenkacz:
Summary:
Right now, for a running job, task status never goes from Starting to Running because once we enter an active state, we stop processing these updates. We only process finished and failed. After this patch, we should propagate the right task status up to the API and UI.

JIRA issues:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,9 @@
 # Next
 
+# 0.6.NEXT
+
+* When querying run detail with `embed=history`, `successfulFinishedRuns` and `failedFinishedRuns` contains new field `tasks` which is an array of taskIds of that finished run. This allow people to query task ids even for finished job runs.
+
 # 0.6.21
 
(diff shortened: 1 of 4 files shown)


changelog.md
```diff
@@ -1,5 +1,9 @@
 # Next
 
+# 0.6.NEXT
```
alenkacz: Oh I see I added changelog here, not the other PR. But nevermind, we're gonna hopefully merge both :)

changelog.md
```diff
@@ -1,5 +1,10 @@
...
+
+* When querying run detail with `embed=history`, `successfulFinishedRuns` and `failedFinishedRuns` contains new field `tasks` which is an array of taskIds of that finished run. This allow people to query task ids even for finished job runs.
```
kensipe: "This allow people" -> "This will allow people"

PR #340, opened 2019-06-12
Fixing SI Provisioning and Launch for Tests

kensipe:
Summary:

Debian 9 is the platform.  jessie-backports is not necessary for the installation of gettext-base.

JIRA issues:   DCOS-48406


kensipe:
```
```
We have got to fix our AMIs

PR #341, opened 2019-06-18
Pin paramiko to 2.4.2 to fix the broken SI tests

takirala:
Summary:

As of shakedown `1.5.0`, paramiko is a moving dependency. Because a newer version of paramiko is released recently (`2.5.0` was released on June 10), it broke the SI test dependencies. Ideally, we shakedown should pin the dependency and we should use the dependency from it. Until such time where we have a released version of shakedown with frozen dependencies, this PR pins the paramiko manually to a last known working version.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -27,7 +27,7 @@ ansiColor('gnome-terminal') {
         } finally {
             junit(allowEmptyResults: true, testResults: 'target/test-reports/*.xml')
             junit(allowEmptyResults: true, testResults: 'tests/integration/target/test-reports/*.xml')
-            archiveArtifacts "*sandboxes.tar.gz"
+            archiveArtifacts artifacts: "*sandboxes.tar.gz", allowEmptyArchive: true
             archiveArtifacts "*log.tar.gz"
         }
       }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -27,8 +27,8 @@ ansiColor('gnome-terminal') {
         } finally {
             junit(allowEmptyResults: true, testResults: 'target/test-reports/*.xml')
             junit(allowEmptyResults: true, testResults: 'tests/integration/target/test-reports/*.xml')
-            archive includes: "*sandboxes.tar.gz"
-            archive includes: "*log.tar.gz"
+            archiveArtifacts "*sandboxes.tar.gz"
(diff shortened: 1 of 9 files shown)


Jenkinsfile
```diff
@@ -27,8 +27,8 @@ ansiColor('gnome-terminal') {
         } finally {
             junit(allowEmptyResults: true, testResults: 'target/test-reports/*.xml')
             junit(allowEmptyResults: true, testResults: 'tests/integration/target/test-reports/*.xml')
-            archive includes: "*sandboxes.tar.gz"
-            archive includes: "*log.tar.gz"
+            archiveArtifacts "*sandboxes.tar.gz"
+            archiveArtifacts "*log.tar.gz"
```
takirala: I have seen the `archive includes:` to be not working. It prints something similar to:
([here is a sample build](https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/system-integration-tests/job/metronome-si-dcos-open/job/master/527/console))
```log
The archive step is deprecated, please use archiveArtifacts instead."
No files found to archive for pattern "diagnostics.zip", continuing.
```

Changing it to archiveArtifacts worked ([proof](https://jenkins.mesosphere.com/service/jenkins/blue/organizations/jenkins/system-integration-tests%2FMetronome%20SI%20pull%20request/detail/Metronome%20SI%20pull%20request/4/artifacts/)) though. I would imagine `deprecated` means the feature should still work but not sure if I am missing something.

tests/system/Makefile
```diff
@@ -15,6 +15,7 @@ help:
 init:
 	pip3 install pipenv
 	pipenv sync
+	pipenv graph
```
takirala: This is just for informative purpose, i ll remove it if you feel its noisy.

tests/system/Pipfile
```diff
@@ -1,22 +1,16 @@
...
-
 [requires]
-
 python_version = "3.6"
 
-
 [packages]
-
-shakedown = {git = "https://github.com/dcos/shakedown.git", ref = "1.5.0"}
+paramiko = "==2.4.2"
```
takirala: All I did was `pipenv install paramiko==2.4.2` and the resulting diff in `Pipfile` and `Pipfile.lock` was a result of that.

kensipe:
I'm so tired of
```
```

PR #343, opened 2019-06-26
Fix metronome not reviving when starting new job

alenkacz:
Summary:
Offer matcher reconciler was not run with `.start()` so the source was emitting no value and `combineLatest` needs at least one value to publish *ANY* result.

We don't need OfferMatcherReconciler at all since we don't use PVs (as pointed out by Matthias).

JIRA issues: DCOS_OSS-5166

### DIFF
--- a/jobs/src/main/scala/dcos/metronome/scheduler/SchedulerModule.scala
+++ b/jobs/src/main/scala/dcos/metronome/scheduler/SchedulerModule.scala
@@ -170,10 +170,7 @@ class SchedulerModule(
   private[this] lazy val offersWanted: Source[Boolean, Cancellable] = {
     offerMatcherManagerModule.globalOfferMatcherWantsOffers
       .via(EnrichedFlow.combineLatest(offerMatcherReconcilerModule.offersWantedObservable, eagerComplete = true))
-      .map { case (managerWantsOffers, reconciliationWantsOffers) => {
-        println(s"Processing offersWanted with ${managerWantsOffers} and ${reconciliationWantsOffers}")
-        managerWantsOffers || reconciliationWantsOffers
-      } }
+      .map { case (managerWantsOffers, reconciliationWantsOffers) => managerWantsOffers || reconciliationWantsOffers }
   }
 
   val launchQueueModule = new LaunchQueueModule(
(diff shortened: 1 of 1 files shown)


jobs/src/main/scala/dcos/metronome/scheduler/SchedulerModule.scala
```diff
@@ -97,9 +97,6 @@ class SchedulerModule(
       persistenceModule.groupRepository,
       leadershipModule)(actorsModule.materializer)
 
-  lazy val taskStatusProcessor: TaskStatusUpdateProcessor = new TaskStatusUpdateProcessorImpl(
```
alenkacz: this is unrelated, it was just unused field

timcharper:
This fixes the issue upstream: https://github.com/mesosphere/marathon/pull/6948

jobs/src/main/scala/dcos/metronome/scheduler/SchedulerModule.scala
```diff
@@ -88,22 +88,9 @@ class SchedulerModule(
...
-      instanceTrackerModule.instanceTracker,
-      persistenceModule.groupRepository,
-      leadershipModule)(actorsModule.materializer)
-
-  lazy val taskStatusProcessor: TaskStatusUpdateProcessor = new TaskStatusUpdateProcessorImpl(
-    metrics, clock, instanceTrackerModule.instanceTracker, schedulerDriverHolder, killService, eventBus)
-
   private[this] lazy val launcherModule: LauncherModule = {
     val instanceTracker: InstanceTracker = instanceTrackerModule.instanceTracker
     val offerMatcher: OfferMatcher = StopOnFirstMatchingOfferMatcher(
```
meichstedt: Take a look at `StopOnFirstMatchingOfferMatcher`: you don't even need to use that one and could just do
```
val offerMatcher: OfferMatcher = offerMatcherManagerModule.globalOfferMatcher
```

:D

alenkacz: LOL :) You're right as always :)

kensipe:
ugh
```
+ bin/install-protobuf.sh
Linux
script returned exit code 100
```

PR #346, opened 2019-07-02
Bump cron-utils to 9.0.0 the latest release

kensipe:
Summary:
There are a number of bugs fixed between 7.0.5 to 9.0.0.   The most significant is the infinite loop bug for midnight DST:  https://github.com/jmrozanec/cron-utils/pull/333

It is confusing why we are still using threeten-bp.  It was originally required for cron-utils but isn't now.   I've completely removed it.   We previously were on 1.3.3 (released in 2016).  There were a number of fixes up to the latest 1.4.0 version https://www.threeten.org/threetenbp/changes-report.html#a1.4.0.   
At this point, it isn't necessary and removes a point of failure to eliminating it.

JIRA issues:


kensipe:
also... did I mention that it is tiring seeing this flake in our CI all the time?

```
+ bin/install-protobuf.sh
Linux
script returned exit code 100```

PR #348, opened 2019-07-03
Update to Release Build Node for CI

kensipe:
Summary:
updating to a later version of build node to try to avoid the apt-get lock issue

JIRA issues:


jeschkies:
Won't do for now.

PR #351, opened 2019-07-30
fix auth for job runs to be constant time

kensipe:
Summary:
Removed unnecessary permission checking for every active run and instead of checking it only once for the `jobSpec`

JIRA issues:  COPS-5143

### DIFF
--- a/api/src/main/scala/dcos/metronome/api/v1/controllers/JobRunController.scala
+++ b/api/src/main/scala/dcos/metronome/api/v1/controllers/JobRunController.scala
@@ -3,7 +3,7 @@ package api.v1.controllers
 
 import dcos.metronome.api.v1.models._
 import dcos.metronome.api.{ ApiConfig, Authorization, UnknownJob, UnknownJobRun }
-import dcos.metronome.jobrun.JobRunService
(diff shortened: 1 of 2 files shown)


meichstedt:
> Fixes timeout issue associated with multi calls to job runs

I tested this manually and it did not fix the issue in Metronome 0.4
The problem does not occur on Metronome master with 1500 JobRuns, so I couldn't confirm the change has any implication wrt COPS-5143

(It does lool like a reasonable improvement though!)

api/src/main/scala/dcos/metronome/api/v1/controllers/JobRunController.scala
```diff
@@ -30,7 +30,20 @@ class JobRunController(
 
   def getJobRuns(id: JobId) = measured {
     AuthorizedAction.async { implicit request =>
-      jobRunService.activeRuns(id).map(_.filter(request.isAllowed)).map(Ok(_))
+      //      jobRunService.activeRuns(id).map(_.filter(request.isAllowed)).map(Ok(_))
```
zen-dog: Let's remove this line

PR #353, opened 2019-08-23
Apply concurrencyPolicy for manually triggered JobRuns as well.

meichstedt:
Summary:
`concurrencyPolicy` is a property on Schedules. Manually triggering a `JobRun` should be considered scheduling related as well, so it seems reasonable to reject manually triggered `JobRun`s when `concurrencyPolicy` is set to `FORBID`.

This is a PoC. To be discussed:
* It's not clear what a `concurrencyPolicy` of `FORBID` on a disabled schedule should mean
* We might want to opt-in to this behavior in the next version, and make it the default in the one after.

This is related to COPS-5204.

### DIFF
--- a/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunServiceActor.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunServiceActor.scala
@@ -3,6 +3,7 @@ package jobrun.impl
 
 import java.time.Clock
 
+import akka.actor.Status.Failure
 import akka.actor._
 import dcos.metronome.eventbus.TaskStateChangedEvent
 import dcos.metronome.jobrun.StartedJobRun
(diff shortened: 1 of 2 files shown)


jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunServiceActor.scala
```diff
@@ -71,7 +72,8 @@ class JobRunServiceActor(
 
     val skipRun = schedule.exists(schedule => schedule.concurrencyPolicy == ConcurrencyPolicy.Forbid && runsForJob(spec.id).nonEmpty)
     if (skipRun) {
-      log.info(s"Skipping scheduled run for ${spec.id} based on concurrency policy")
+      log.warning(s"Skipping scheduled run for ${spec.id} based on concurrency policy")
+      sender() ! Failure(ConcurrentJobRunNotAllowed(spec))
```
meichstedt: Funny enough, `ConcurrentJobRunNotAllowed` existed since June 2016, but was never used because we didn't reject manual triggers …

meichstedt:
I actually wanted to make this a draft – it's missing tests and discussion.

jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunServiceActor.scala
zen-dog: 👏

kensipe:
Because policies like Concurrency were defined in the schedule... I took them to be only applied for that schedule and not for manual triggers.    When we reason that there are multiple schedules even though we only allow 1 and we couldn't get consensus on getting rid of multiple schedules, it wasn't possible to know which policy to use.   this design only works IMO if we agree that only 1 schedule is ever possible.   leaving open the option of refactoring a lot of code out of the code base.

kensipe:
@meichstedt what do we want to do here?

PR #356, opened 2019-10-01
Update to marathon 1.9

ANeumann82:
First approach. Intercepts the InstanceChanges and updates the goal before delivering finished instances to LaunchQueue, etc. 

The problem here is, if we fail over at certain points, the launch queue might get access to the intercepted states and relaunch an already finished task

JIRA issues: DCOS-50471


kensipe:
this is exciting stuff... nice work!!

PR #363, opened 2019-10-25
Fix zkStore locations

ANeumann82:
Old Metronome versions did not store job-specs in /state/job-specs but /job-specs.
We need to do this in the Curator-Version as well to be able to migrate correctly


ANeumann82:
I'm not sure we need a new jira for this, it is actually a fix for the change of MARATHON-8709, I didn't test it good enough and introduced the changed zk-path.

PR #366, opened 2020-03-09
Add networking support to Metronome

timcharper:


jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -14,7 +14,7 @@ import mesosphere.marathon.{ MarathonSchedulerDriverHolder, StoreCommandFailedEx
 import mesosphere.marathon.core.launchqueue.LaunchQueue
 import mesosphere.marathon.core.task.Task
 import org.apache.zookeeper.KeeperException.NodeExistsException
-import scala.async.Async.{ async, await }
+import scala.async.Async.async
```
jeschkies, line 17: This is odd. When would one use `async` without `await`?

timcharper: Yeah... good point lol. This was scalafix remove unused imports effect.

PR #373, opened 2020-05-29
Load instances asynchronously.

jeschkies:
Summary:
The `JobRunExecutorActor` has a few blocking calls that can eat up the
thread pool and block the instance track from replying to calls. This
change removes a few blocking calls.

JIRA issues: COPS-6174

### DIFF
--- a/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
@@ -404,8 +404,8 @@ object JobRunExecutorActor {
     new JobRunExecutorActor(run, promise, persistenceActorRefFactory,
       launchQueue, instanceTracker, driverHolder, clock))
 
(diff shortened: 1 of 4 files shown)


jobs/src/main/scala/dcos/metronome/jobrun/impl/JobRunExecutorActor.scala
```diff
@@ -55,17 +55,20 @@ class JobRunExecutorActor(
...
-            context.become(starting)
-          case _ =>
-            self ! Initialized(Nil)
+        async {
+          val existingTasks = await(tasksFromInstanceTracker())
+          val alreadyQueued = await(existsInLaunchQueue())
+          existingTasks match {
+            case tasks if tasks.nonEmpty =>
+              // the actor was probably just restarted and we did not lose contents of the launch queue
+              self ! Initialized(tasks)
```
timcharper, line 64: I think in reality this is fine; having async code access the actor context is error prone, however. We should ideally switch this to pipeTo. I'd also prefer to factor out the async logic to an object method so it does not have access to the internal actor state.

timcharper: I extracted out the async block to an object method

PR #374, opened 2020-06-11
Migrate build to Jenkins on DC/OS.

jeschkies:
Summary:
This migrates the build onto DC/OS and should reduce build cost and
issues.

This patch also replaces the HTTP status matcher with one that supplies
more debugging information on failure.


timcharper:
Is this ready for review ? Did we get the large Jenkins instance size?

jeschkies:
> Is this ready for review ?

https://github.com/mesosphere/usi/pull/156 needs to land first but this is ready for review.

> Did we get the large Jenkins instance size?

No, but I don't want to wait.

ci/set_port_range.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+# See http://mesos.apache.org/documentation/latest/port-mapping-isolator/
```
pierrebeitz, line 3: 404 for me. 
found https://mesos.readthedocs.io/en/latest/port-mapping-isolator/

ci/set_port_range.sh
```diff
@@ -0,0 +1,7 @@
...
+# and mesosphere.util.PortAllocator docs.
+echo "Ephemeral port range before: $(cat /proc/sys/net/ipv4/ip_local_port_range)"
+sysctl -w net.ipv4.ip_local_port_range="60001 61000"
+echo "Ephemeral port range after: $(cat /proc/sys/net/ipv4/ip_local_port_range)"
```
pierrebeitz, line 7: why do we need to do that?

tests/integration/src/test/scala/dcos/metronome/integrationtest/utils/MetronomeFramework.scala
```diff
@@ -142,6 +142,9 @@ object MetronomeFramework {
         }
     }(collection.breakOut)
 
+    lazy val url = s"http://localhost:$httpPort"
+    lazy val facade = new MetronomeFacade(url)
```
pierrebeitz, line 146: where are those used?

Jenkinsfile
```diff
@@ -4,31 +4,61 @@
...
+    stage('Test') {
+      agent {
+        node {
+          label 'docker'
+        }
+      }
+      steps {
+        ansiColor('xterm') {
+          sh 'sudo ci/set_port_range.sh'
+          sh 'sudo sbt integration/test || true'
```
pierrebeitz, line 54: what's the `|| true` for? we don't want to propagate a failure of integration-tests?

ci/set_port_range.sh
jeschkies: Good find. Seems the link is outdated. This is a copy from Marathon.

jeschkies:
```suggestion
# See https://mesos.readthedocs.io/en/latest/port-mapping-isolator/
```

ci/set_port_range.sh
jeschkies: This relates to https://github.com/mesosphere/usi/blob/master/test-utils/src/main/scala/com/mesosphere/utils/PortAllocator.scala#L11. Alex found that some of our parallel integration tests failed because ports would be assigned twice and thus services wouldn't be able to bind to a certain port. Unfortunately this is the only way to tell Mesos a port range.

Jenkinsfile
jeschkies: Yes. Not for now. The tests are too flaky until https://github.com/dcos/metronome/pull/375 merges.

tests/integration/src/test/scala/dcos/metronome/integrationtest/utils/MetronomeFramework.scala
jeschkies: See line 40 in tests/integration/src/test/scala/dcos/metronome/integrationtest/MetronomeITBase.scala.

PR #377, opened 2020-06-25
Format code with scalafmt.

jeschkies:
Summary:
USI and Marathon are formatted the same way. This diff is just unifying
the formatting.


pierrebeitz:
@jeschkies i again got a different result by running scalafmt in the root folder and added an according commit. feel free to remove that again if it does not make sense to you.

PR #378, opened 2020-07-13
Fold schedules into job spec model.

jeschkies:
Summary:
This patch allows to specify schedules when creating job specifications.
Schedules are always embedded in job spec queries.

JIRA issues: DCOS_OSS-5970

### DIFF
--- a/api/src/main/resources/public/api/v1/schema/jobspec.schema.json
+++ b/api/src/main/resources/public/api/v1/schema/jobspec.schema.json
@@ -27,10 +27,7 @@
       }
     },
     "schedules": {
-      "type": "array",
-      "items": {
-        "$ref": "schedulespec.schema.json"
-      }
+      "type": "array"
     },
     "run": {
       "type": "object",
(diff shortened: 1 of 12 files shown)


api/src/main/resources/public/api/v1/schema/jobspec.schema.json
```diff
@@ -26,6 +26,9 @@
         "type": "string"
       }
     },
+    "schedules": {
+      "type": "array"
```
jeschkies, line 30: I thought we could add
```
"items": { "$ref": "schedulespec.schema.json" }
```

However, our validation does not support these kind of references.

timcharper: Darn; does a newer version support it?

If not, can we at least copy-pasta?

jeschkies: > Darn; does a newer version support it?

I don't quite understand the Metronome implementation. I don't think the JSON schema expects schemas to be loaded from custom paths but via URIs.

> If not, can we at least copy-pasta?
I'd say so.

api/src/main/scala/dcos/metronome/api/Binders.scala
```diff
@@ -49,6 +49,7 @@ object Binders {
         params: Map[String, scala.collection.Seq[String]]
     ): Option[Either[String, Set[Embed]]] = {
       val embeds = params.getOrElse(key, Seq.empty).flatMap(_.split(","))
+      // TODO: marks schedules as deprecated.
```
jeschkies, line 52:
```suggestion
```

PR #379, opened 2020-07-15
Define dependencies on job spec.

jeschkies:
Summary:
This patch introduces the `dependencies` field to job specs. Job specs
cannot define dependencies on parents.

JIRA issues: DCOS_OSS-5971

### DIFF
--- a/jobs/src/main/scala/dcos/metronome/jobspec/impl/JobSpecServiceDelegate.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobspec/impl/JobSpecServiceDelegate.scala
@@ -46,5 +46,8 @@ class JobSpecServiceDelegate(config: JobSpecConfig, actorRef: ActorRef, metrics:
       actorRef.ask(DeleteJobSpec(id)).mapTo[JobSpec]
     }
 
-  override def transaction(updater: Seq[JobSpec] => Option[Modification]): Future[JobSpec] = ???
+  override def transaction(updater: Seq[JobSpec] => Option[Modification]): Future[Option[JobSpec]] =
+    updateJobSpecTimeMetric {
+      actorRef.ask(Transaction(updater)).mapTo[Option[JobSpec]]
+    }
 }
--- a/jobs/src/main/scala/dcos/metronome/jobspec/impl/JobSpecServiceDelegate.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobspec/impl/JobSpecServiceDelegate.scala
@@ -4,6 +4,7 @@ package jobspec.impl
 import akka.actor.ActorRef
 import akka.pattern.ask
 import akka.util.Timeout
+import com.mesosphere.usi.async.ExecutionContexts
 import dcos.metronome.jobspec.impl.JobSpecServiceActor._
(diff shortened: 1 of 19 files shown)


jobs/src/main/scala/dcos/metronome/model/JobSpec.scala
```diff
@@ -29,14 +33,36 @@ case class JobSpec(
...
+
+  def acyclic[JobId]: Validator[Seq[JobId]] = new NullSafeValidator[Seq[JobId]](
+    test = _ => false, // TODO: check if the dependencies are cyclic.
+    failure = _ -> "is not acyclic"
+  )
+
+  def validJobSpec(jobSpecService: JobSpecService): Validator[JobSpec] = validator[JobSpec] { jobSpec =>
+
+    def exist: Validator[Seq[JobId]] = new NullSafeValidator[Seq[JobId]](
+      test = ids => ids.exists(id => Await.result(jobSpecService.getJobSpec(id), ???).isEmpty),
```
jeschkies, line 56: I really do not want to block here. So I think we should move this into the controller.

jobs/src/main/scala/dcos/metronome/model/JobSpec.scala
```diff
@@ -29,14 +33,36 @@ case class JobSpec(
 
 object JobSpec {
   val DefaultDescription = None
+  val DefaultDependencies = Option.empty[Seq[JobId]]
```
jeschkies, line 36: The job ids would require to join each time. It's simpler to deserialize though.

jobs/src/main/scala/dcos/metronome/model/JobSpec.scala
```diff
@@ -29,14 +33,74 @@ case class JobSpec(
...
+    val unknownDependencies = jobSpec.dependencies.toSet -- directDependencies
+    if (unknownDependencies.nonEmpty) {
+      throw ValidationError(
+        s"Dependencies contain unknown jobs. unknown=[${unknownDependencies.mkString(", ")}]"
+      )
+    }
+
+    // Validate that graph is acyclic
+
+    /* Assumption: The current state allSpecs has now cycles.
```
jeschkies, line 74:
```suggestion
    /* Assumption: The current state allSpecs has no cycles.
```

jobs/src/main/scala/dcos/metronome/jobspec/impl/JobSpecServiceDelegate.scala
```diff
@@ -45,4 +46,9 @@ class JobSpecServiceDelegate(config: JobSpecConfig, actorRef: ActorRef, metrics:
     deleteJobSpecTimeMetric {
       actorRef.ask(DeleteJobSpec(id)).mapTo[JobSpec]
     }
+
+  override def transaction(updater: Seq[JobSpec] => Option[Modification]): Future[Option[JobSpec]] =
+    updateJobSpecTimeMetric {
+      actorRef.ask(Transaction(updater)).mapTo[JobSpec].map(Option.apply)(ExecutionContexts.callerThread)
```
jeschkies, line 52: We don't always get a job spec back.

PR #382, opened 2020-07-29
Execute dependent job when all dependencies finished.

jeschkies:
Summary:
This the business logic of MIP 0008 "Jobs with Dependencies". We had an actor analog
to the `JobSpecSchedulerActor`. This new actor subscribes to all job run events and updates
an internal state. Once all parents have a successful run that is new than the child run we
trigger new job execution.

JIRA issues: DCOS_OSS-5974

### DIFF
--- a/jobs/src/main/scala/dcos/metronome/jobspec/impl/JobSpecDependencyActor.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobspec/impl/JobSpecDependencyActor.scala
@@ -1,0 +1,48 @@
+package dcos.metronome.jobspec.impl
+
+import akka.actor.{Actor, ActorLogging, Props, Stash}
+import dcos.metronome.jobrun.JobRunService
+import dcos.metronome.model.{Event, JobSpec}
+
+/**
+  * Manages one JobSpec.
+  *
+  * If the JobSpec has dependencies, it subscribes to the events of its dependencies, ie parents.
+  *
+  * This actor is analog to [[JobSpecSchedulerActor]].
+  *
+  * TODO: find a better name
+  */
+class JobSpecDependencyActor(initSpec: JobSpec, runService: JobRunService) extends Actor with Stash with ActorLogging {
+
+  val dependencyIndex = initSpec.dependencies.toSet
+
+  override def preStart(): Unit = {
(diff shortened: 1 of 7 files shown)


jobs/src/main/scala/dcos/metronome/jobspec/impl/JobSpecDependencyActor.scala
```diff
@@ -0,0 +1,134 @@
...
+
+  override def receive: Receive = idling
+
+  def idling: Receive = {
+    case ev: Event.JobRunEvent =>
+      dependenciesState.update(ev.jobRun)
+      if (dependenciesState.shouldTriggerJob(lastSuccessfulRun)) {
+        log.info(s"Start next run of job ${spec.id}, all parents finished successfully")
+        runService.startJobRun(initSpec).pipeTo(self)
+        context.become(running)
```
jeschkies, line 51: I told @pierrebeitz already what limitation we have with this.

First, I think we need to handle the result of `startJobRun`. It stands to question whether we need another receive method for that.

Second and this is more important, I saw job runs triggered each time there was an update. Imagine C run five seconds depends on A and B. B is run every ten seconds but A every ten millis. This would trigger C multiple times.

A way to avoid this is to save the `lastCreatedAt` for C and use this as well to check whether we already scheduled a run for C or not. I will create a follow up as a proposal.

jobs/src/main/scala/dcos/metronome/jobspec/impl/JobSpecDependencyActor.scala
```diff
@@ -0,0 +1,134 @@
...
+  * If the JobSpec has dependencies, it subscribes to the events of its dependencies, ie parents.
+  *
+  * This actor is analog to [[JobSpecSchedulerActor]].
+  *
+  * TODO: find a better name
+  */
+class JobSpecDependencyActor(initSpec: JobSpec, runService: JobRunService) extends Actor with Stash with ActorLogging {
+
+  private[impl] var spec = initSpec
+  private[impl] var lastSuccessfulRun: Instant = Instant.MIN
```
jeschkies, line 28: Should load the last job run as a failover pattern?

PR #385, opened 2020-10-06
D2IQ-71940 don't let cyclic deps crash the validation

pierrebeitz:
we enhance the DFS here to only inspect nodes that have not been
inspected before.
the actor is grateful for it does not need to check the same dependencies over and over again for all eternity if someone tries to introduce a cyclic dependency.

## JIRAs

https://jira.d2iq.com/browse/D2IQ-71940


pierrebeitz:
@timcharper, @kaiwalyajoshi, @ANeumann82 can someone please have a look here?

timcharper (approved):

PR #386, opened 2020-10-13
MARATHON-8772 use the latest spec to run dependants

pierrebeitz:
formerly we were using `initSpec` whenever we ran a dependany job. now
each dependencyActor also listens to `JobSpecUpdated` in order to update
its local copy of its jobSpec.

i'm really sure there is a more elegant solution to this. additionally
i'm not sure why we'd need an actor for managing dependencies at all.
but all that does not matter here, as we want to fix this by a given
deadline.

a major downside of this commit is that i'm not fluent enough in this
codebase to quickly assemble an according test.

closes MARATHON-8772

### DIFF
--- a/jobs/src/main/scala/dcos/metronome/jobspec/impl/JobSpecDependencyActor.scala
+++ b/jobs/src/main/scala/dcos/metronome/jobspec/impl/JobSpecDependencyActor.scala
@@ -34,6 +34,8 @@ class JobSpecDependencyActor(initSpec: JobSpec, runService: JobRunService) exten
   // An index of all parents
   private[impl] var dependencyIndex: Set[JobId] = spec.dependencies.toSet
 
+  private[impl] var currentSpec: JobSpec = initSpec
+
   // State of latest parents' runs.
   val lastParentRuns: mutable.Map[JobId, JobRun] = mutable.Map.empty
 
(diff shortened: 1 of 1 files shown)


timcharper (approved):

jobs/src/main/scala/dcos/metronome/jobspec/impl/JobSpecDependencyActor.scala
```diff
@@ -78,10 +81,13 @@ class JobSpecDependencyActor(initSpec: JobSpec, runService: JobRunService) exten
         log.info(
           s"Start next run of job ${spec.id}, all parents finished successfully lastParentRuns=${lastParentRuns.values}"
         )
-        runService.startJobRun(initSpec).pipeTo(self)
+        runService.startJobRun(currentSpec).pipeTo(self)
         context.become(starting)
       }
 
+    case ev: Event.JobSpecUpdated =>
```
timcharper, line 88: Any idea what happens when the job is deleted? Does this come across as an updated event?

Should we terminate the actor in that case?

pierrebeitz: it was my take that this is handled [here](https://github.com/dcos/metronome/blob/b7b4a2c7569592a22fa4d7e1f06a92330df4d30a/jobs/src/main/scala/dcos/metronome/jobspec/impl/JobSpecServiceActor.scala#L167-L179).

